PR #4, opened 2020-06-13
Exporter Landing page

roidelapluie:


web/web.go
```diff
@@ -0,0 +1,40 @@
...
+package web
+
+import (
+	"image/color"
+	"net/http"
+)
+
+// Config represents the configuration of the web listener.
+type Config struct {
+	Color       color.Color // Used for the landing page header
```
brian-brazil, line 23: Does this need to be configurable?

web/web.go
```diff
@@ -0,0 +1,40 @@
...
+	Name        string      // The name of the exporter, generally suffixed by _exporter
```
brian-brazil, line 24: Full stops

web/web.go
roidelapluie: Thanks for the very early review. I think that having just a different color for each exporter would be a nice 'touch'; without going to the point where we offer themes.

SuperQ: Yes, it's nice idea. For example, using the MySQL brand colors for the mysqld_exporter. :smile:

SuperQ:
Ping, want to get this basic version checked in? I would like to add to it.

roidelapluie:
Hello, at this stage we still need CSS. You should have pisg access to the branch if you want to help I think.

web/generated_assets.go
```diff
@@ -0,0 +1,37 @@
...
+            {{ range .Links }}
+            <li><a href="{{ .Address }}">{{.Text}}</a>{{ if .Description }}: {{.Description}}{{end}}</li>
+            {{ end }}
+            </ul>
+        </nav>
+    </body>
+</html>
+`
+var cssContent = `Hello
+`
```
SuperQ, line 37: How about something simple to start.

```suggestion
var cssContent = `
h1 {
   color: #fff;
   background-color: #343a40
}
`
```

SuperQ:
Ping, it'd be nice to get the basics in so we can continue to build on it.

SuperQ:
Closing in favor of https://github.com/prometheus/exporter-toolkit/pull/51

PR #12, opened 2020-11-16
Add kingping integration

roidelapluie:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,5 +7,6 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/common v0.15.0
 	golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9
+	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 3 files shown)


https/tls_config.go
```diff
@@ -164,8 +166,23 @@ func ConfigToTLSConfig(c *TLSStruct) (*tls.Config, error) {
 	return cfg, nil
 }
 
-// Listen starts the server on the given address. If tlsConfigPath isn't empty the server connection will be started using TLS.
+// AddFlags adds the flags used by this package to the Kingpin application.
+// To use the default Kingpin application, call AddFlags(kingpin.CommandLine)
+func AddFlags(a *kingpin.Application) {
```
SuperQ, line 171: Let's call this `AddKingpinFlags`, as we could include other flag library support in the future.

https/tls_config.go
```diff
@@ -164,8 +166,23 @@ func ConfigToTLSConfig(c *TLSStruct) (*tls.Config, error) {
...
+	a.Flag(
+		"web.config",
+		"Path to config yaml file that can enable TLS or authentication.",
+	).Default("").StringVar(&cliWebConfigFlag)
+}
+
+// Listen starts the server on the given address. If tlsConfigPath or
```
SuperQ, line 178: > If tlsConfigPath or tlsConfigPath aren't empty

Wat?

roidelapluie:
@SuperQ WDYT?

https/tls_config.go
```diff
@@ -164,14 +162,21 @@ func ConfigToTLSConfig(c *TLSStruct) (*tls.Config, error) {
 	return cfg, nil
 }
 
-// Listen starts the server on the given address. If tlsConfigPath isn't empty the server connection will be started using TLS.
-func Listen(server *http.Server, tlsConfigPath string, logger log.Logger) error {
-	if tlsConfigPath == "" {
+// TLSConfig contains the configuration required to start an tls/basic auth server.
```
SuperQ, line 165: I'd like this to be more clear that it's a path to a yaml file.

Do we need to get complex enough that this is a struct?

Maybe just something like this:

```
type TLSConfigFilePath string
```

roidelapluie:
@SuperQ WDYT now?

SuperQ (approved):
Simple and easy, I like it.

PR #17, opened 2020-12-25
Return 401 when bad credentials are passed

roidelapluie:
https://tools.ietf.org/html/rfc7235#section-3.1

> If the request included authentication credentials, then the 401
> response indicates that authorization has been refused for those
> credentials.

cc @SuperQ 


SuperQ (approved):
LGTM

PR #18, opened 2020-12-25
http: Fix linting

roidelapluie:
circleci was not enabled for forks, I have changed the settings.

cc @SuperQ 


roidelapluie:
Let's do this in #17

PR #20, opened 2020-12-25
Add experimental flag

roidelapluie:

### DIFF
--- a/https/kingpinflag/flag.go
+++ b/https/kingpinflag/flag.go
@@ -20,7 +20,7 @@ import (
 // To use the default Kingpin application, call AddFlags(kingpin.CommandLine)
 func AddFlags(a *kingpin.Application) *string {
 	return a.Flag(
-		"web.config",
-		"Path to config yaml file that can enable TLS or authentication.",
+		"web.config.file",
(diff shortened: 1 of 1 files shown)


https/kingpinflag/flag.go
```diff
@@ -20,7 +20,16 @@ import (
 // To use the default Kingpin application, call AddFlags(kingpin.CommandLine)
 func AddFlags(a *kingpin.Application) *string {
 	return a.Flag(
-		"web.config",
-		"Path to config yaml file that can enable TLS or authentication.",
+		"web.config.file",
```
SuperQ, line 23: Do we want to change this? I don't want to break existing node_exporter users by changing the flag name.

roidelapluie: #21 will break node_exporter users :(

Note exporter can keep its current flag as hidden.

SuperQ: Yea, good point, we can keep a compatibility flag alias.

SuperQ (approved):

roidelapluie:
Actually we should make it experimental for everyone, and not have a dedicated function I think.

roidelapluie:
I even think that http.config.file would be better

SuperQ (approved):
LGTM

PR #21, opened 2020-12-25
Make certificates and CA relative to config file

roidelapluie:
cc @SuperQ 


roidelapluie:
This can break node_exporter but this is the behaviour of Prometheus server.

Note that the feature is marked as experimental in node_exporter.

SuperQ (approved):
LGTM

PR #24, opened 2020-12-26
Basic auth: add metrics

roidelapluie:

### DIFF
--- a/https/users.go
+++ b/https/users.go
@@ -48,8 +48,7 @@ func (u *userAuthRoundtrip) instrument(r prometheus.Registerer) {
 	u.failuresCounter = prometheus.NewCounter(
 		prometheus.CounterOpts{
 			Namespace: "prometheus_toolkit",
-			Subsystem: "https",
-			Name:      "request_basic_authentication_failures_total",
+			Name:      "http_request_basic_authentication_failures_total",
 			Help:      "Total number of requests rejected by basic authentication because of wrong username, password, or configuration.",
 		},
 	)
--- a/https/users.go
+++ b/https/users.go
@@ -17,6 +17,7 @@ import (
 	"net/http"
 
 	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/crypto/bcrypt"
 )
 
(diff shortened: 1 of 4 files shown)


https/users.go
```diff
@@ -37,14 +38,29 @@ func validateUsers(configPath string) error {
...
+	handler         http.Handler
+	logger          log.Logger
+	failuresCounter prometheus.Counter
+}
+
+func (u *userAuthRoundtrip) instrument(r prometheus.Registerer) {
+	u.failuresCounter = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "prometheus_toolkit_web",
+			Name:      "http_request_basic_authentication_failures_total",
```
SuperQ, line 51: Maybe match the namespace and subsystem:

```suggestion
			Namespace: "prometheus_toolkit",
			Subsystem: "https",
			Name:      "request_basic_authentication_failures_total",
```

roidelapluie: nice, done!

roidelapluie:
Ready for a second review, once that is in I will add metrics for the TLS part (or I can do in the same PR if you want)

https/users.go
```diff
@@ -37,14 +38,30 @@ func validateUsers(configPath string) error {
...
+	handler         http.Handler
+	logger          log.Logger
+	failuresCounter prometheus.Counter
+}
+
+func (u *userAuthRoundtrip) instrument(r prometheus.Registerer) {
+	u.failuresCounter = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "prometheus_toolkit",
+			Subsystem: "https",
```
brian-brazil, line 51: https and basic auth are not the same thing

In general, it's better to have the metric name as a single string literal as it makes grepping and determining what the actual metric name is easier.

roidelapluie: I have renamed the metric.  However I think that keeping namespace separated is better, it that's fine for you.

SuperQ: I preferred it named based on the package name.

roidelapluie: Should we change the package name?

roidelapluie:
Would renaming the package to http help, like this?

roidelapluie:
I did not like the "http" package rename as it is too close with go's net/http.

I have reworked this to have a metric name: `prometheus_toolkit_https_basic_authentication_failures_total` which does not read as 'https_requests' anymore. does it sound better ?

ArthurSens:
Hey hey, closing due to inactivity, but please open again if there's still work to be done here :)

PR #25, opened 2020-12-26
Add CodeQL

roidelapluie:


roidelapluie:
I would expect he workflow to run on this PR, for some reason it does not

PR #32, opened 2021-01-13
Cache basic authentications

roidelapluie:
In order to improve the response time with basic authentication, cache
the results of bcrypt in memory.

This is a draft to see if that approach would be acceptable.

The code is coming from Caddy, and I have fullfiled the attribution
required by Apache-2 by adding the copyright to our headers:
https://github.com/caddyserver/caddy/commit/9a7756c6e4b4ddb945bede3ddb2dfbf241208915

### DIFF
--- a/web/cache.go
+++ b/web/cache.go
@@ -1,0 +1,77 @@
+// Copyright 2021 The Prometheus Authors
+// This code is partly borrowed from Caddy:
+//    Copyright 2015 Matthew Holt and The Caddy Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package web
+
+import (
+	weakrand "math/rand"
+	"sync"
+	"time"
+)
+
+var cacheSize = 100
+
+func init() {
+	weakrand.Seed(time.Now().UnixNano())
+}
+
+type cache struct {
+	cache map[string]bool
+	mtx   sync.Mutex
+}
+
+// newCache returns a cache that contains a mapping of plaintext passwords
+// to their hashes (with random eviction). This can greatly improve the
+// performance of traffic-heavy servers that use secure password hashing
+// algorithms, with the downside that plaintext passwords will be stored in
+// memory for a longer time (this should not be a problem as long as your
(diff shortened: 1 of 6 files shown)


roidelapluie:
If the general approach is file I will add specific tests.

web/users.go
```diff
@@ -58,7 +62,21 @@ func (u *userAuthRoundtrip) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	user, pass, auth := r.BasicAuth()
 	if auth {
 		if hashedPassword, ok := c.Users[user]; ok {
-			if err := bcrypt.CompareHashAndPassword([]byte(hashedPassword), []byte(pass)); err == nil {
+			cacheKey := hex.EncodeToString(append(append([]byte(user), []byte(hashedPassword)...), []byte(pass)...))
```
roidelapluie, line 65: We could also sha512 this one maybe if we don't want plaintext in memory.

web/cache.go
```diff
@@ -0,0 +1,77 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package web
+
+import (
+	weakrand "math/rand"
```
SuperQ, line 19: Odd choice to go with this over `crypto/rand`.

roidelapluie: It comes from caddy and I guess that we do not need more, it is just used for eviction, not for anything security related.

SuperQ: Yea, I understand, I'm just interested in the choice.

web/users.go
SuperQ: Basic auth is plain text anyway, so I'm not that concerned about it.

web/cache.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2020 The Prometheus Authors
```
SuperQ, line 1:
```suggestion
// Copyright 2021 The Prometheus Authors
```

roidelapluie: fixed

roidelapluie:
is this good to go?

roidelapluie:
Hold on.

roidelapluie:
OK now it works as expected. I have added getters and setters to release locks early.

I have also added a Lock around bcrypt so that we only run one at the same time.

SuperQ (approved):
LGTM

web/users.go
```diff
@@ -40,6 +44,10 @@ type userAuthRoundtrip struct {
 	tlsConfigPath string
 	handler       http.Handler
 	logger        log.Logger
+	cache         *cache
+	// bcryptMtx is there to ensure that bcrypt.CompareHashAndPassword is run
+	// only once in parallel as this is CPU expansive.
```
beorn7, line 49: expansive → intensive?

roidelapluie: 🤦🏻 I will fix it

PR #34, opened 2021-01-16
Release v0.5.1

roidelapluie:
@SuperQ @beorn7 

Now that #32 is merged, the plan is to release a bugfix release of Prometheus with this.

I'd like to cut a v0.5.1 release of the toolkit in such a way that I can resume my work on #24, which would end up in a next release of the toolkit but not in the prometheus 2.24 release.


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.5.1 / 2021-01-15
+
+* [IMPROVEMENT] Cache basic authentication results to significantly improve
```
beorn7, line 3: Don't we usually call this `[ENHANCEMENT]`?

roidelapluie: Yes we do :)

roidelapluie: Also should I write `performances` or `performance`?

beorn7: I'd say `performance` is better.

`performances` sounds like a number of dance performances or something. :-)

roidelapluie: 👯

beorn7 (approved):

roidelapluie:
We still need #39 so we do not introduce a side-channel timing vulnerability.

SuperQ (approved):

PR #39, opened 2021-01-18
Prevent user enumeration

roidelapluie:
This commit makes sure that the timing is similar when a user exists or
does not exist.


SuperQ (approved):
LGTM

PR #51, opened 2021-04-16
Exporter Landing page

SuperQ:
Add a basic exporter landing page.

### DIFF
--- a/web/landing_page.css
+++ b/web/landing_page.css
@@ -1,0 +1,13 @@
+body {
+  font-family: -apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica Neue,Arial,Noto Sans,Liberation Sans,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol,Noto Color Emoji;
+  margin: 0;
+}
+nav {
+  background-color: {{.HeaderColor}};
+  color: #fff;
+  font-size: 2rem;
+  padding: 1rem;
+}
+.main {
+  padding: 1rem;
+}
--- a/web/landing_page.go
+++ b/web/landing_page.go
@@ -1,0 +1,82 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !genassets
+// +build !genassets
+
+//go:generate go run -tags genassets gen_assets.go
+
+package web
+
+import (
+	"bytes"
+	_ "embed"
+	"net/http"
+	"text/template"
+)
+
+// Config represents the configuration of the web listener.
+type LandingConfig struct {
+	HeaderColor string         // Used for the landing page header.
+	CSS         string         // CSS style tag for the landing page.
+	Name        string         // The name of the exporter, generally suffixed by _exporter.
+	Description string         // A short description about the exporter.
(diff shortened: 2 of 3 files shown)


web/web.go
```diff
@@ -0,0 +1,42 @@
...
+	config LandingConfig
+}
+
+func NewHandler(c LandingConfig) *handler {
+	return &handler{
+		config: c,
+	}
+}
+
+func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
roidelapluie, line 41: You should serve the files here

SuperQ: Yup, I know, I'm going to work on finishing this soon. I'm probably going to more directly integrate this into the https handler code.

LeviHarrison:
Might be nice to be able to "highlight" a few important metrics so that the user doesn't have to go digging in the `/metrics` page (eg. the Memcached exporter has the memory usage displayed).

SuperQ:
It's still really ugly, but it works.

image

SuperQ:
After this, I can add some more of the boiler plate stuff automatic in the web and kingpinflag packages.

web/generated_assets.go
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !genassets
+// +build !genassets
+
+package web
```
Nexucis, line 17: maybe it would be worth to add a comment explaining this file is generated and shouldn't be modified.

Nexucis: like 
`// THIS FILE HAS BEEN GENERATED, DO NOT EDIT.`

Nexucis:
> @LeviHarrison That would be interesting, we could possibly pull a `prometheus.Describe()` to get the metric TYPE/HELP list.

if at some point we would like something a bit less home made, we should create a simple React app. That might help also to have something less ugly ;)

web/generated_assets.go
SuperQ: Yup, that's a good idea.

SuperQ:
Yup, I would love to have something less hand-rolled. But I don't want to get _too_ fancy. This is just a simple exporter landing page. I would like to keep it as a single HTML page if possible.

Nexucis:
yeah that's the risk to go with React, it won't be that simple anymore and as a side effect potentially the binary size increase more than what we would like for a simple page.

Nexucis:
so It's a trade off.
It's just if at some point we want that involved javascript for example we should go with React. Until that, we should be good with what you are doing @SuperQ

Nexucis (approved):

SuperQ:
Yea, I want to avoid JS if possible. It _might_ be useful for some exporters that have forms for validation. Like blackbox_exporter. But exporter endpoints are meant to be minimal by design. I don't want to hear about any security issues for this feature.

roidelapluie (commented):
I will review this shortly. We should avoid the generate assets and use embed.

SuperQ:
Oh, I wasn't aware of `embed`. I'll update to that, looks way easier.

SuperQ:
Oh, I wasn't aware of `embed`. I'll update to that, looks way easier.

SuperQ:
Ok, I made some improvements to the look.

image

Nexucis:
> We should avoid the generate assets and use embed.

Ah yeah I forgot about that everytime.

Nexucis:
looks definitely better with `embed` !!

dswarbrick:
> Ok, I made some improvements to the look.

Can we maybe embed some minimal CSS to use a sans-serif font, so that it looks a bit more modern?

dswarbrick:
> It's just if at some point we want that involved javascript for example we should go with React.

Please keep in mind the difficulty of packaging some transpiled web scripts, especially in Debian, which has quite strict policies regarding that. For example, the Alertmanager web UI is already completely stripped in the Debian prometheus-alertmanager package, due to its gargantuan list of dependencies that would need to be packaged also.

I see no real justification to bloat a simple splash page with scripts.

SuperQ:
@dswarbrick How about this:

image

PR #57, opened 2021-07-05
Minor grammar and typo fixes in web-config docs

neomantra:


docs/web-configuration.md
```diff
@@ -86,8 +86,8 @@ which will look something like:
 `$2y$10$X0h1gDsPszWURQaxFh.zoubFi6DXncSjhoQNJgRrnGs7EsimhC7zG`
 
 The cost (10 in the example) influences the time it takes for computing the
-hash. A higher cost will en up slowing down the authentication process.
-Depending on the machine, a cost of 10 will take about ~70ms where a cost of
+hash. A higher cost will end up slowing down the authentication process.
+Depending on the machine, a cost of 10 will take about ~70ms, whereas a cost of
 18 can take up to a few seconds. That hash will be computed on every
```
roidelapluie, line 91:
```suggestion
18 can take up to a few seconds. That hash will be computed on the first authenticated HTTP request and then cached.
```

roidelapluie (commented):
Thanks, while you are at it, would you mind fixing a small error we have?

Thanks!!

docs/web-configuration.md
neomantra: Sure, I just force-pushed with that change.

roidelapluie: I do not see the commit

neomantra: Problem was between keyboard and chair ;) should be there now!

roidelapluie (approved):
Thanks!

PR #58, opened 2021-07-09
Fix ClientAuth type typo

LeviHarrison:
`RequireAnyClientCert` is the actual name, not `RequireClientCert`

Fixes prometheus/prometheus#9066


LeviHarrison:
Forgot to `git add` my tests 🤣

roidelapluie (approved):

LeviHarrison:
We'll need to do a new release for this

LeviHarrison:
Which brings up a question I've had, which is why is the `TLSConfig` even in here, it would make a lot more sense to have it in common.

PR #59, opened 2021-07-12
Release 0.6.1

roidelapluie:


roidelapluie:
I actually think we should drop VerifyClientCertIfGiven RequireAnyClientCert and RequestClientCert.

LeviHarrison:
Ehhhhhhh we refer to https://golang.org/pkg/crypto/tls/#ClientAuthType for the types so I think we should support all of them.

LeviHarrison:
I guess we could move that documentation into our own repo and eliminate those options

roidelapluie:
So, should we go ahead with this? We support all of them but users should be aware that only one is secure.

LeviHarrison:
Sure

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.1 / 2021-06-30
+
+* [BUGFIX] Allow RequireAnyClientCert as client_auth_type.
```
LeviHarrison, line 3: Should we add the PR number?

LeviHarrison (approved):
Looks good

PR #61, opened 2021-08-20
Fix client TLS certificate tests

WGH-:
First, in lots of existing tests, `client_ca_file` was set to the CA of the _server_. However, it's actually used to verify the client certificate, but there're no tests where the client presents any.

In the second commit, I add a couple of tests where the client actually presents a certificate.


WGH-:
> Hello, this is now conflicting, could you please rebase? Thanks!

Done! Sorry that I couldn't get around rebasing earlier, considering that there weren't any major conflicts

PR #62, opened 2021-08-20
Support for client certificate fingerprint whitelist

WGH-:
This is useful for simple setups that don't require full-blown CAs for mutual authentication.

An example of a program that support this type of client auth is Postfix (see [`permit_tls_clientcerts`](http://www.postfix.org/postconf.5.html#permit_tls_clientcerts)).

This introduces a footgun in a way: using `client_fingerprints` practially requires using `RequireAnyClientCert`, so it invites a mistake of setting `RequireAnyClientCert` and forgetting about `client_fingerprints`. Luckily, it's possible to discern empty list from a nil one (not set), but it still requires the user to carefully replace multiline YAML list with `client_fingerprints: []` if he wants to delete all certificates for some reason. Maybe that's not really a problem, but I still want to mention it.

Alternatively, we might introduce a new custom `client_auth_type` (say, `FingerprintWhitelist`) that would behave like `RequireAnyClientCert`, but would treat a nil set like an empty one (i.e. reject any certificate). This diverges a bit from `tls.ClientAuthType` enumeration list, but it's less error-prone this way.

Currently based on top of #61, hence draft.


WGH-:
> Wouldn't putting client cert as CA work instead? This looks uncommon.

I think not, because it might lack the flag allowing it to be a CA.

ArthurSens:
Hey, I'm taking the liberty of closing this PR since there hasn't been any activity in a while. If there's still work to be done here, please let us know!

PR #63, opened 2021-09-07
Users: Log unauthorized requests

networkException:
This patch adds a new configuration option to the web.config.file which
makes it possible to enable logging of unauthorized requests.

If "log_unauthorized" at config file's top level is set to true,
any unauthorized request will have the ip as well as the X-Forwarded-For
header logged. This way, a program that might parse the logs can determent
for itself if the X-Forwarded-For header can be trusted.

### DIFF
--- a/web/users.go
+++ b/web/users.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"golang.org/x/crypto/bcrypt"
 )
 
@@ -41,10 +42,12 @@ func validateUsers(configPath string) error {
 }
 
 type userAuthRoundtrip struct {
-	tlsConfigPath string
-	handler       http.Handler
-	logger        log.Logger
-	cache         *cache
+	tlsConfigPath     string
+	handler           http.Handler
+	logger            log.Logger
+	requestLogger     log.Logger
+	requestLoggerLock sync.RWMutex
+	cache             *cache
 	// bcryptMtx is there to ensure that bcrypt.CompareHashAndPassword is run
 	// only once in parallel as this is CPU intensive.
 	bcryptMtx sync.Mutex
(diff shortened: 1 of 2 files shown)


web/users.go
```diff
@@ -93,6 +94,10 @@ func (u *userAuthRoundtrip) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if c.LogUnauthorized {
+		level.Info(u.logger).Log("msg", "Unauthorized", "ip", r.RemoteAddr, "forwarded", r.Header.Get("X-Forwarded-For"))
```
SuperQ, line 98: Maybe make the forwarded for header configurable?

SuperQ:
Seems like a good idea to me.

web/users.go
networkException: Do you mean *if* the header should be used as the (only?) output or *what* header should be printed?

SuperQ: I was thinking something like this:
```yaml
log_unauthorized:
  remote_ip_header: "X-Forwarded-For"
```

This would allow users to customize which `X-Forwarded-For` header to use (I've seen some proxies/CDNs use non-standard headers :shrug:)

networkException: Good idea, I think I will just log one field "ip" then and use a header for that if specified in the config

roidelapluie:
Can we log to a file? We might also log successful queries, with usernames?

networkException:
I think logging to a file is a bit out of scope of this pr, usually there's always a way to pipe the output somewhere. If people want it I can also implement logging on success

roidelapluie:
We should not mix the two streams together (application / access logs). it is separate things.

roidelapluie:
prometheus also uses json for logging the queries.

networkException:
I have rewritten the implementation to output to a json file

SuperQ:
I would prefer to avoid multiple log streams. The fact that Prometheus logs queries to a separate file is an anti-pattern for operations.

We separate logging types in our tools like ELK and Loki.

networkException:
For my usecase it doesn't matter, I'd just like to have some way of outputting failed requests. I'm not able to decide what design would be best for the project but I'll happily implement a different one if that gets agreed on.

SuperQ:
@networkException Yea, let me have a discussion with @roidelapluie about this policy. I really want to avoid a proliferation of log files in the Prometheus ecosystem. It makes dealing with deployments a lot more work.

roidelapluie:
You can put /dev/stdout if you don't want a log file

ArthurSens:
Hi @SuperQ @roidelapluie , have you two come to an agreement? :)

PR #70, opened 2021-10-31
Allow auth to be skipped for some paths

darshanime:
closes https://github.com/prometheus/prometheus/issues/9166
Signed-off-by: darshanime <deathbullet@gmail.com>


roidelapluie (commented):
Thanks for your contribution!

This should be documented, and we should use the configuration we read on every request.

I think a better name for it would be `basic_auth_excluded_paths`

Issue #71, opened 2021-11-05
CORS requests are failed if Basic Auth is enabled


stokito:
I enabled CORS headers and basic auth but a browser still can't fetch information from Prometheus.
Before sending a GET request an OPTIONS request if performed first and it doesn't contains the Authorization header.
So Prometheus must allow the OPTIONS request without the Authorization header.
I can do that with some [Nginx configuration](https://gist.github.com/tomkersten/1323477) but I don't want to use just for this and also I want to keep configuration clear.

roidelapluie:
You are correct, we should allow OPTIONS without basic auth, I will move this to the exporter-toolkit repository. We probably do not want this by default, but as an option in web.yml.

roidelapluie:
I guess we will also need to understand how Prometheus itself handles OPTIONS requests (without basic auth).

stokito:
I created a PR. Honestly I don't see benefits of having this configurable. In theory, that may lead to problem when someone tested an API endpoint with the OPTIONS covered by a basic auth and then they may be wondered that someone disabled basic auth and now sensitive data is exposed.
Th OPTIONS was created for as a replacement for Flash crossdomain.xml but originally the method was introduced in WebDAV. For a security reason it must not be a resource specific but rather show features of the server itself.
As a possible risk that I see it may be in some API endpoints be a check `if r.Method == GET {} else { POST }` i.e. that it always assume that there is only two possible http methods and the 405 error is not returned properly.
There is no that many api endpoints to check

roidelapluie:
```
curl -s -X OPTIONS https://node.demo.do.prometheus.io/metrics |head 
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 4.141e-05
go_gc_duration_seconds{quantile="0.25"} 5.7685e-05
go_gc_duration_seconds{quantile="0.5"} 6.1472e-05
go_gc_duration_seconds{quantile="0.75"} 6.9851e-05
go_gc_duration_seconds{quantile="1"} 0.000358368
go_gc_duration_seconds_sum 297.390302862
go_gc_duration_seconds_count 4.113281e+06
# HELP go_goroutines Number of goroutines that currently exist.
```

```
curl -s -X WHATEVER https://node.demo.do.prometheus.io/metrics |head 
[...]

PR #79, opened 2022-01-28
#71 Skip basic auth for OPTIONS http method

stokito:
The OPTIONS needed for CORS requests


roidelapluie:
While I understand the need, I need to assess the security impact, as said in the issue. Not only for Prometheus but for generic exporters as well.

roidelapluie:
```
curl -s -X OPTIONS https://node.demo.do.prometheus.io/metrics |head 
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 4.141e-05
go_gc_duration_seconds{quantile="0.25"} 5.7685e-05
go_gc_duration_seconds{quantile="0.5"} 6.1472e-05
go_gc_duration_seconds{quantile="0.75"} 6.9851e-05
go_gc_duration_seconds{quantile="1"} 0.000358368
go_gc_duration_seconds_sum 297.390302862
go_gc_duration_seconds_count 4.113281e+06
# HELP go_goroutines Number of goroutines that currently exist.
```

```
curl -s -X WHATEVER https://node.demo.do.prometheus.io/metrics |head 
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 4.141e-05
go_gc_duration_seconds{quantile="0.25"} 5.7685e-05
go_gc_duration_seconds{quantile="0.5"} 6.1472e-05
go_gc_duration_seconds{quantile="0.75"} 6.9851e-05
go_gc_duration_seconds{quantile="1"} 0.000358368
go_gc_duration_seconds_sum 297.390426157
go_gc_duration_seconds_count 4.113283e+06
# HELP go_goroutines Number of goroutines that currently exist.
```

It looks like the Prometheus golang API does not check the method. This fix would then disable basic auth for metrics endpoints if OPTIONS is used as method.

stokito:
I recently switched to Go and when I first time saw the HTTP handler API my first thought was "hey, here is not forced to check a method like in Java servlet api". Also all code examples even from go tour didn't handled this :(

SuperQ:
@stokito I agree, we should be much more careful about request method handling here and in other parts of the http handler(s).

web/handler.go
```diff
@@ -97,7 +97,8 @@ func (u *webHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set(k, v)
 	}
 
-	if len(c.Users) == 0 {
+	// If OPTIONS method or none users configured then skip auth check
+	if r.Method == http.MethodOptions || len(c.Users) == 0 {
```
SuperQ, line 101: One possible safe way to do this would be to check both the method and the URI.

```suggestion
	if (r.RequestURI == "/" && r.Method == http.MethodOptions) || len(c.Users) == 0 {
```

This way we only allow bypass for the typical browser [Preflight request](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing#Preflight_example).

stokito: the additional is not needed. Or maybe I miss something?

roidelapluie:
This is not safe. Instead we should try to reply ourselves when there is an OPTIONS query.

It would mean moving CORS config into the toolkit.

gagbo:
Hello,

> I will try to detail what I have in mind if it's not clear.

Don't you think that checking for `OPTIONS` should be the responsibility of the exporter that's being wrapped by the toolkit? The fact that prometheus has [its own logic to set CORS headers](https://github.com/prometheus/prometheus/blob/main/util/httputil/cors.go) makes me think so.

When you say that you want to move CORS config in the toolkit, that would mean moving all this code, and the CORS origin regexp config value, into the toolkit? If so, that'd be a breaking change I assume, do you have a timeline for this change, or could some help make such a change come faster and available in prometheus API?

PR #80, opened 2022-01-29
Smaller fixes and changes

mrueg:
Some smaller cleanups and updates.
* Update dependencies
* Test against more recent versions of golang
* Drop outdated fallback code


roidelapluie (commented):
Thanks for your pull request. The golint workflow file is propagated from prometheus/prometheus and can not be changed in this repository. Additionally, the TLS tests should not cover all the cases, just enough to be working as expected, so  I don't think we should make them version dependent.

mrueg:
Thanks for the quick feedback, I've dropped the golangci-lint change and hope the change will make it work on go1.16 as well. 

I changed the order of the output for bad cipher selection, please check if my assumption there was correct.

Unfortunately  for the `valid tls config yml with multiple client ciphers` it seems like there's a difference in preference since go 1.17:
`AES-128 is preferred over AES-256 for encryption.`  https://go.dev/blog/tls-cipher-suites (so it picks a different one than with go 1.16)

I can pick other ciphers, unless you have a different idea?

roidelapluie:
Yes, let's pick other ciphers for testing.

SuperQ (approved):
LGTM

roidelapluie (approved):

roidelapluie:
Thank you!

PR #84, opened 2022-04-26
fix CVE CVE-2022-27191 in x/crypto module in go.mod package

pgvishnuram:
Our CVE scanner detected a HIGH CVE caused by x/crypto module with ID CVE-2022-27191

issue link: https://github.com/prometheus/exporter-toolkit/issues/85
Signed-off-by: pgvishnuram <vishnu@astronomer.io>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/go-kit/log v0.2.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/common v0.32.1
-	golang.org/x/crypto v0.0.0-20220128200615-198e4374d7ed
+	golang.org/x/crypto v0.0.0-20220315160706-3147a52a75dd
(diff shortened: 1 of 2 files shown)


SuperQ:
The linked CVE is about the ssh package. This library does not use this code, your vulnerability report is a false positive.

Please do not report raw vulnerability scanner results. They are prone to false positives and cause the Prometheus team toil in verifying. Please verify vulnerability reports and include specific details as to which components are directly exploitable. Please also include a reproduction case.

pgvishnuram:
see the go.mod its still using vulnerable package version, if its not at all being used then we should remove them from go.mod

pgvishnuram:
you are using "golang.org/x/crypto/bcrypt" this is from golang.org/x/crypto package

SuperQ:
Yes, but the vulnerability is about `golang.org/x/crypto/ssh`, which is not in use. Your vulnerability scanner is defective and reporting matches incorrectly.

PR #97, opened 2022-06-06
Further secure TLS communications

### LINKED ISSUE #96: Further secure TLS communications
Currently when using TLS, the servers will accept requests from any client that has a certificate signed by the specified Certificate Authority. As such, I'd like to see custom server certificate validation supported. This will help enforce deny-by-default.

I'd like to be able to pass a flag, such as `-cert-allowed-cn`, that can be used to create a custom `VerifyPeerCertificate` (part of the crypto/tls package) and can be passed as a callback directly to the tls config. All this function needs to do is verify that the seen common-name is the same as the expected common-name.

Willing to submit a PR if the maintainers think this is a good idea. Thanks!

saroshali-dbx:
Closes https://github.com/prometheus/exporter-toolkit/issues/96

### DIFF
--- a/web/testdata/web_config.auth_client_common_name.good.yaml
+++ b/web/testdata/web_config.auth_client_common_name.good.yaml
@@ -1,0 +1,6 @@
+tls_server_config:
+  cert_file: "server.crt"
+  key_file: "server.key"
+  client_auth_type: "RequireAndVerifyClientCert"
+  client_ca_file: "client_selfsigned.pem",
+  client_cert_allowed_cn: "prometheus.example.com"
--- a/web/testdata/web_config_auth_client_common_name.bad.yaml
+++ b/web/testdata/web_config_auth_client_common_name.bad.yaml
@@ -3,4 +3,4 @@ tls_server_config:
   key_file: "server.key"
   client_auth_type: "RequireAndVerifyClientCert"
   client_ca_file: "client_selfsigned.pem"
-  client_cert_allowed_cn: "bad.example.com"
+  client_cert_allowed_cn: "bad"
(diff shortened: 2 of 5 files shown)


saroshali-dbx:
gentle ping @roidelapluie @SuperQ :) 

For context, here's the issue created for the downstream dependency - https://github.com/weaveworks/common/issues/242

SuperQ:
I thought most certificate validation has moved to verifying SANs, not CNs.

saroshali-dbx:
> I thought most certificate validation has moved to verifying SANs, not CNs.

Yeah, thats the intention with SANs to be able to get around the limitation of only being able to specify a single server-name in CN. I ported this functionality from etcd -- and my organization still utilizes common-name -- but I can update this PR to check SANs first and then fallback to CNs. How does that sound?

roidelapluie:
Actually that's a good point @SuperQ , even go uses SAN's and has removed support for CN. I think we should follow go and use SANs

SuperQ (changes requested):
Common Name is deprecated for validation. This needs to use SANs.

From [RFC 2818, May 2000](https://www.ietf.org/rfc/rfc2818.html):

>  Although the use of the Common Name is existing practice, it is deprecated and Certification Authorities are encouraged to use the dNSName instead.

Chrome-based browsers dropped CN-only in 2017.

logopk:
Just a question: is SAN or any other DNS unusual for client certs? I usually create them with 

`C = DE, ST = xxx, L = xx, O = xxx, OU = xxx, CN=me@myself.de`

Thus given the 3.2 
`SHOULD check the identity as described above`

leads me to
`Otherwise, the (most specific) Common Name field in the Subject field of the certificate MUST be used`

ArthurSens:
Thank you for the work on tightening client-certificate validation and for the thoughtful discussion around Common Names and SANs. The validation direction raised in review was not resolved here, and the TLS implementation has moved on enough that this branch no longer applies cleanly. We are closing it as part of the backlog cleanup, but the use case and discussion remain valuable references. If the need remains, we would welcome a fresh PR based on the current TLS implementation.

Issue #85, opened 2022-04-26
fix CVE-2022-27191 in x/crypto package


pgvishnuram:
Our CVE scanner identified a HIGH CVE in kube-state-metrics service, which can be fixed by updating the x/crypto package
![image](https://user-images.githubusercontent.com/81585115/165243530-a6e5ade3-b315-4fd3-b2dc-ecb7b3a4d86f.png)

SuperQ:
Please do not report raw vulnerability scanner results. They are prone to false positives and cause the Prometheus team toil in verifying. Please verify vulnerability reports and include specific details as to which components are directly exploitable. Please also include a reproduction case.

PR #103, opened 2022-07-31
Bump golang.org/x/crypto to v0.0.0-20220315160706-3147a52a75dd

amosricky:
For CVE-2022-27191, the golang.org/x/crypto/ssh package before 0.0.0-20220314234659-1baeb1ce4c0b for Go allows an attacker to crash a server in certain circumstances involving AddHostKey.

Although the module golang.org/x/crypto/ssh is not used in the project, it still be scanned and reported for every raw vulnerability scanner. Maybe we could just upgrade it to safe version to avoid the unnecessary discussion like [this](https://github.com/prometheus/exporter-toolkit/issues/85)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/go-kit/log v0.2.1
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/common v0.35.0
-	golang.org/x/crypto v0.0.0-20220128200615-198e4374d7ed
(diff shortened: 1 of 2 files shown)


SuperQ:
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

amosricky:
> This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

Fixed.

PR #105, opened 2022-08-10
Fix tests with Go 1.18

### LINKED ISSUE #83: Tests fail with Go 1.18 due to denial of SHA1 certificates
This bug is present with commit hash 9f7b277231e81a5529f560c0389d764cc0ac7de3 (latest master as of this writing).

Per the Go 1.18 release notes: "crypto/x509 will now reject certificates signed with the SHA-1 hash function. This doesn't apply to self-signed root certificates. Practical attacks against SHA-1 [have been demonstrated since 2017](https://shattered.io/) and publicly trusted Certificate Authorities have not issued SHA-1 certificates since 2015."

This is causing quite a few of the test cases to fail with Go 1.18.

simonpasquier:
`go test ./...` failed with Go 1.18 because the server certificate and
key used SHA1. To fix this, the server certificate/key (and certificate
authority chain) have been copied from the github.com/prometheus/common
repository.

Additionally one test failed because the error message returned by the
stdlib has been updated in Go 1.18. The test has been modified to
support both variants.

Signed-off-by: Simon Pasquier <spasquie@redhat.com>

Closes #83


SuperQ (approved):
LGTM

PR #107, opened 2022-08-15
web: Replace pkg/errors with stdlib errors

mrueg:
Since https://github.com/pkg/errors is archived by the owner, let's replace is with stdlib features.


SuperQ (approved):
LGTM, Thanks

PR #115, opened 2022-10-16
Update golangci linters

mrueg:
Replace deprecated golint with revive

Add goimports and misspell linters as they are used in https://github.com/prometheus/prometheus/blob/main/.golangci.yml


SuperQ (approved):
Thanks!

PR #120, opened 2022-11-15
Cleaning up repeated addresses before listening

knabben:
in a particular corner case of the windows_exporter (https://github.com/prometheus-community/windows_exporter/issues/1092), the values of `weblistenaddress` come duplicated, making the host listen to twice the same value when using configuration.

One possible option is to flatten and remove duplicates before starting the socket for the list of addresses. this can be shortened since the flag library has not been released since 2017.

golang 1.9 should bring this function, for now copying it over until the bump happens


SuperQ:
> golang 1.9 should bring this function, for now copying it over until the bump happens

Do you mean 1.19?

web/tls_config.go
```diff
@@ -226,6 +227,24 @@ func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) e
 	return ServeMultiple(listeners, server, flags, logger)
 }
 
+// compact replaces consecutive runs of equal elements with a single copy.
```
SuperQ, line 230: Hmm, shouldn't we fully unique the slice rather than just consecutive duplicates?

knabben: yes, maybe converting to a set is better here. The question is more, does it makes sense to dedup on this part of the code?

SuperQ (approved):
Thanks

roidelapluie:
Why does this happen in the first place? Will the default bind address always be included then???

roidelapluie:
I think the exporter should clean the struct before passing it to the toolkit, and it should not be part of the toolkit. Feels too risky.

knabben:
closing this PR, agree it should be fixed in the root of the issue instead.

PR #126, opened 2023-01-10
Client TLS: Add option to require a specific DNS name in Subject Alternate Name

### LINKED ISSUE #96: Further secure TLS communications

chodges15:
This is a follow up to the work done in #97 which addresses the requested change by @roidelapluie and @SuperQ to go from CN to SAN verification. Closes #96.

Command used to update the certs can be found in the commit message.


chodges15:
@roidelapluie and @SuperQ this is ready for review when you get a chance, I'd like to get this into Loki for tighter security

chodges15:
@SuperQ - can you kick off CI for this PR again? Linter error that was causing previous error should be resolved.

web/tls_config.go
```diff
@@ -51,6 +51,7 @@ type TLSConfig struct {
 	MinVersion               TLSVersion `yaml:"min_version"`
 	MaxVersion               TLSVersion `yaml:"max_version"`
 	PreferServerCipherSuites bool       `yaml:"prefer_server_cipher_suites"`
+	ClientCertAllowedSanDNS  string     `yaml:"client_cert_allowed_san_dns"`
```
SuperQ, line 54: Should we future proof this a bit and make it a list of allowed SANs? Or maybe an allow regex?

web/tls_config.go
```diff
@@ -51,6 +52,8 @@ type TLSConfig struct {
 	MinVersion               TLSVersion `yaml:"min_version"`
 	MaxVersion               TLSVersion `yaml:"max_version"`
 	PreferServerCipherSuites bool       `yaml:"prefer_server_cipher_suites"`
+	// regular expression to match the SAN DNS entries of the client cert
+	ClientCertAllowedSanDNSRegex string `yaml:"client_cert_allowed_san_dns"`
```
SuperQ, line 56: Minor nit, SANs can be a variety of things, not just DNS names. Let's just simplify this.

How about this:

```suggestion
	ClientAllowedSanRegex string `yaml:"client_allowed_san_regex"`
```

web/tls_config.go
chodges15: good idea, ended up going with an allow regex that checks against all SAN fields

chodges15:
@SuperQ - can I get another round of review for this?

SuperQ:
Looks good, can you update the README docs?

VERSION
```diff
@@ -1 +1 @@
-0.8.1
+0.8.2
```
SuperQ, line 1: Please don't touch the VERSION file.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.2 / 2023-02-28
```
SuperQ, line 1: This should be marked as unreleased

```suggestion
## master / unreleased
```

chodges15:
@SuperQ - docs updated, ready for another review

SuperQ (changes requested):
Minor nit, otherwise LGTM.

SuperQ (approved):
LGTM

roidelapluie:
please remove the changelog

chodges15:
@roidelapluie - changelog removed.

roidelapluie:
I think we should force the regex to be anchored, as it is common in Prometheus.

web/tls_config.go
```diff
@@ -66,6 +68,75 @@ func (t *TLSConfig) SetDirectory(dir string) {
...
+}
+
+// Regexp encapsulates a regexp.Regexp and makes it YAML marshalable.
+type Regexp struct {
+	*regexp.Regexp
+}
+
+// NewRegexp creates a new anchored Regexp and returns an error if the
+// passed-in regular expression does not compile.
+func NewRegexp(s string) (Regexp, error) {
```
chodges15, line 106: This struct borrowed from [Prometheus](https://github.com/prometheus/prometheus/blob/edc8e75c402dd3a4231344c19452645efd7b6209/model/relabel/relabel.go#L160-L165)
If needed, this can be refactored by moving into `github.com/prometheus/common`, though it would be nice if that could be done in a follow up PR.

docs/web-configuration.md
```diff
@@ -37,6 +37,16 @@ tls_server_config:
...
+  
+  # Verify that the client certificate has a Subject Alternate Name (SAN) which includes the following
+  # regex pattern, else terminate connection. SAN match can be one or multiple of the following:
+  # DNS, IP, e-mail, or URI address from https://pkg.go.dev/crypto/x509#Certificate.
+  [ client_allowed_san_regex: <string> | default ""]
+
+  # Verify that the client certificate has a Subject Alternate Name (SAN) which includes the following
+  # regex pattern, else terminate connection. SAN match can be one or multiple of the following:
+  # DNS, IP, e-mail, or URI address from https://pkg.go.dev/crypto/x509#Certificate.
+  [ client_allowed_san_regex: <string> | default ""]
```
roidelapluie, line 49: This is documented twice

chodges15:
@roidelapluie - documentation fixed, not sure how I missed that.

web/tls_config.go
```diff
@@ -163,6 +234,11 @@ func ConfigToTLSConfig(c *TLSConfig) (*tls.Config, error) {
 		cfg.ClientCAs = clientCAPool
 	}
 
+	if c.ClientAllowedSanRegex.Regexp != nil {
+		// verify that the client cert contains the allowed domain name
+		cfg.VerifyPeerCertificate = c.VerifyPeerCertificate
```
roidelapluie, line 239: Why do we need that if?

web/tls_config.go
```diff
@@ -66,6 +68,75 @@ func (t *TLSConfig) SetDirectory(dir string) {
...
+	// sender cert comes first, see https://www.rfc-editor.org/rfc/rfc5246#section-7.4.2
+	cert, err := x509.ParseCertificate(rawCerts[0])
+	if err != nil {
+		return fmt.Errorf("error parsing client certificate: %s", err)
+	}
+
+	// Build up a slice of strings with all Subject Alternate Name values
+	sanValues := append(cert.DNSNames, cert.EmailAddresses...)
+
+	for _, ip := range cert.IPAddresses {
```
roidelapluie, line 82: This is not a SAN value

web/tls_config.go
```diff
@@ -66,6 +68,75 @@ func (t *TLSConfig) SetDirectory(dir string) {
...
+		sanValues = append(sanValues, ip.String())
+	}
+
+	for _, uri := range cert.URIs {
```
roidelapluie, line 86: This is not a SAN value

web/tls_config.go
```diff
@@ -66,6 +68,75 @@ func (t *TLSConfig) SetDirectory(dir string) {
...
+// VerifyPeerCertificate will check the SAN entries of the client cert if there is configuration for it
+func (t *TLSConfig) VerifyPeerCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+	// sender cert comes first, see https://www.rfc-editor.org/rfc/rfc5246#section-7.4.2
+	cert, err := x509.ParseCertificate(rawCerts[0])
+	if err != nil {
+		return fmt.Errorf("error parsing client certificate: %s", err)
+	}
+
+	// Build up a slice of strings with all Subject Alternate Name values
+	sanValues := append(cert.DNSNames, cert.EmailAddresses...)
```
roidelapluie, line 80: Email address is not a SAN value

roidelapluie (commented):
Sorry, but I think  we should use upstream go VerifyOptions

web/tls_config.go
chodges15: My reference here was that both [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280) and [Go x509 source](https://github.com/golang/go/blob/a8ca653892a470daf065b7669061d6ce1b4ff2e9/src/crypto/x509/x509.go#L753-L759) indicate that DNS and URIs are valid Subject Alternate Names.

chodges15:
The upstream x509.VerifyHostname does not check all SAN fields, and for fields that it does check it does not support regex, only a wildcard. That being said, if we only care about DNS and IP addresses here (which would handle my use case) I'm fine going with the standard upstream verification.

web/tls_config.go
chodges15: I guess I'm not fully tracking the question, without the `if` the behavior would be to always require the client to verify SAN; unless you are thinking we could default the regex to be fully permissive with something like `.*`?

roidelapluie:
Okay. You are correct about SAN's. I think we can  move ahead, maybe we should explicltly call out in the documentation that dots must be escaped due to the nature of regexes, but I am also happy about removing support for regexes and working with an explicit list of sans instead.

chodges15:
An explicit list feels better here from a simplicity point of view - and more secure to avoid user regex mistakes.

SuperQ:
Sorry about the regex distraction. It seemed like a good idea at the time. :)

roidelapluie:
One CI lint issue.

roidelapluie (approved):

roidelapluie:
thanks!

SuperQ:
Nice!

chodges15:
Thanks for the good feedback @roidelapluie and @SuperQ.

PR #127, opened 2023-02-17
Fix Nil pointer references for WebSystemdSocket

ethan-gallant:
Noticed this when working with https://github.com/Lusitaniae/apache_exporter

If you don't provide a boolean value for WebSystemdSocket (default false) calling ListenAndServe causes the application to SigSegv. This checks if it is nil before evaluating the boolean value.

### DIFF
--- a/web/tls_config.go
+++ b/web/tls_config.go
@@ -33,6 +33,7 @@ import (
 
 var (
 	errNoTLSConfig = errors.New("TLS config is not present")
+	ErrNoListeners = errors.New("no web listen address or systemd socket flag specified")
 )
 
 type Config struct {
@@ -203,7 +204,11 @@ func ServeMultiple(listeners []net.Listener, server *http.Server, flags *FlagCon
 // WebSystemdSocket in the FlagConfig is true. The FlagConfig is also passed on
 // to ServeMultiple.
 func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) error {
-	if *flags.WebSystemdSocket {
+	if flags.WebSystemdSocket == nil && (flags.WebListenAddresses == nil || len(*flags.WebListenAddresses) == 0) {
+		return ErrNoListeners
+	}
+
+	if flags.WebSystemdSocket != nil && *flags.WebSystemdSocket {
 		level.Info(logger).Log("msg", "Listening on systemd activated listeners instead of port listeners.")
 		listeners, err := activation.Listeners()
 		if err != nil {
@@ -214,6 +219,7 @@ func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) e
 		}
 		return ServeMultiple(listeners, server, flags, logger)
 	}
+
 	listeners := make([]net.Listener, 0, len(*flags.WebListenAddresses))
 	for _, address := range *flags.WebListenAddresses {
 		listener, err := net.Listen("tcp", address)


web/tls_config.go
```diff
@@ -214,6 +214,11 @@ func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) e
 		}
 		return ServeMultiple(listeners, server, flags, logger)
 	}
+
+	if flags.WebListenAddresses == nil {
+		*flags.WebListenAddresses = []string{":9090"}
+	}
```
SuperQ, line 220: We definitely should not do this, as some users do not want any web listeners, only systemd socket.

SuperQ (changes requested):
Maybe a better check would be to make sure that there is at least one valid listener and error if there is none.

web/tls_config.go
dswarbrick: I agree, in addition to it being highly presumptuous for a library (as opposed to an app) to choose a default listener address.

roidelapluie: +1

ethan-gallant: I definitely agree, I've implemented an error log instead. Cheers! :)

web/tls_config.go
```diff
@@ -214,6 +214,11 @@ func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) e
 		}
 		return ServeMultiple(listeners, server, flags, logger)
 	}
+
+	if flags.WebListenAddresses == nil {
+		return errors.New("no valid WebListenAddress specified")
```
roidelapluie, line 219: Can we remove the if? WebListenAddress is not required

SuperQ (approved):
Perfect, thanks!

ethan-gallant:
@SuperQ should be good to go. Squashed commits and amended to have the signature

web/tls_config.go
```diff
@@ -203,7 +203,11 @@ func ServeMultiple(listeners []net.Listener, server *http.Server, flags *FlagCon
 // WebSystemdSocket in the FlagConfig is true. The FlagConfig is also passed on
 // to ServeMultiple.
 func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) error {
-	if *flags.WebSystemdSocket {
+	if flags.WebSystemdSocket == nil && (flags.WebListenAddresses == nil || len(*flags.WebListenAddresses) == 0) {
+		return errors.New("no web listen address or systemd socket flag specified")
```
SuperQ, line 207: Minor nit, I like to make these kinds of errors package vars so they can be used with `errors.Is()`.

Something like this:
```go
var ErrNoListeners = errors.New("...")
```

ethan-gallant: I did a constant at the top if you're able to re-review :)

web/tls_config.go
```diff
@@ -35,6 +35,8 @@ var (
 	errNoTLSConfig = errors.New("TLS config is not present")
 )
 
+const ErrNoListeners = errors.New("no web listen address or systemd socket flag specified")
```
SuperQ, line 38: Huh, I didn't know that this was a valid expression. Is this a new Go feature?

SuperQ: Looks like it's still not valid.

```
web/tls_config.go:38:24: errors.New("no web listen address or systemd socket flag specified") (value of type error) is not constant
```

ethan-gallant: I just mashed it out in my IDE but didn't actually test it. I was running go 1.20 so it's possible it's a new thing.

I'll change it to a Var tomorrow.

ethan-gallant: Should be good to go now @SuperQ

SuperQ (approved):

roidelapluie:
Thanks!

Issue #108, opened 2022-08-29
Race condition in tests


NightTsarina:
Hi,

As reported in [Debian bug #1013578](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1013578), the tests fail sometimes with CPU contention, and reliably in single-CPU systems. After some investigation, I have found that the problem is a race condition between `ListenAndServe` and attempting to connect to the http server, as there is no synchronisation in the code.

I have produced a simple patch that adds a few `time.Sleep` calls after starting the server, which seems to reliably fix the problem.

```patch
--- a/web/handler_test.go
+++ b/web/handler_test.go
@@ -18,6 +18,7 @@
 	"net/http"
 	"sync"
 	"testing"
+	"time"
 )
 
 // TestBasicAuthCache validates that the cache is working by calling a password
@@ -42,6 +43,7 @@
 		ListenAndServe(server, "testdata/web_config_users_noTLS.good.yml", testlogger)
 		close(done)
 	}()
+	time.Sleep(250 * time.Millisecond)
 
 	login := func(username, password string, code int) {
 		client := &http.Client{}
@@ -106,6 +108,7 @@
 		ListenAndServe(server, "testdata/web_config_users_noTLS.good.yml", testlogger)
 		close(done)
 	}()
+	time.Sleep(250 * time.Millisecond)
 
 	login := func() {
 		client := &http.Client{}
@@ -150,6 +153,7 @@
 		ListenAndServe(server, "testdata/web_config_headers.good.yml", testlogger)
 		close(done)
 	}()
+	time.Sleep(250 * time.Millisecond)
 
 	client := &http.Client{}
 	req, err := http.NewRequest("GET", "http://localhost"+port, nil)
```

roidelapluie:
See #139

SuperQ:
Fixed!

PR #137, opened 2023-03-07
Fix flakey tests

SuperQ:
Add a small delay between starting the listener and running tests to avoid flakey test race conditions.

Fixes: https://github.com/prometheus/exporter-toolkit/issues/108

### DIFF
--- a/web/handler_test.go
+++ b/web/handler_test.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"sync"
 	"testing"
+	"time"
 )
 
 // TestBasicAuthCache validates that the cache is working by calling a password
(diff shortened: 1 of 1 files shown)


Nexucis:
mmm my experience in this kind of delay is : there is always a moment it's not enough and the tests will fail.

Don't we have a condition we can loop on to know if it's ready ?

SuperQ:
Yes, we could poll the port to see if it's open, with a small delay loop. I thought about doing that, but went with the simple workaround.

roidelapluie:
https://github.com/prometheus/exporter-toolkit/pull/139/files

SuperQ:
Closing in favor of #139

PR #139, opened 2023-03-07
test: use port polling to wait for server to start

roidelapluie:
Previously tests did not wait for the server to start listening on the port. This could lead to race conditions and flaky tests. This commit adds a port polling loop that waits for the server to start listening on the port before running the tests.


SuperQ (approved):
Thanks!

PR #141, opened 2023-03-09
Use Reader interface to read TLS keys and certs

fayzal-g:
Introduce a `SecretReader` interface to fetch secrets (keys/certs) used to configure TLS. Now allows the ability to not just read keys/certs from a file, but via anything that implements `ReadSecret` - for example if one wanted to read keys and certs from Vault instead.

If no interface is passed, then it will default to reading from a file as before, so there should be no change to any existing behaviour.

### DIFF
--- a/web/tls_config.go
+++ b/web/tls_config.go
@@ -36,6 +36,16 @@ var (
 	ErrNoListeners = errors.New("no web listen address or systemd socket flag specified")
 )
 
+type SecretReader interface {
+	ReadSecret(path string) ([]byte, error)
+}
+
+type fileReader struct{}
+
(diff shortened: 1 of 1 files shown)


roidelapluie:
I apologize, but I cannot approve your pull request as it appears to be related to a topic already on the agenda for the Prometheus developer summit. Specifically, there is ongoing discussion about including support for secret providers in Prometheus/Alertmanager, as seen in the following GitHub issues: prometheus/alertmanager#3108 and prometheus/prometheus#11477.

bboreham:
Next Prometheus dev summit is 23rd March 2023, right?

roidelapluie:
I will not merge it before we work on the new secret manager

PR #145, opened 2023-03-25
Refer user to doco from help message

DRuggeri:
The help message generated by the toolkit leaves users with no hints about what to put in the `web.config.file`. Because this toolkit is designed to be embedded in other projects, it would be non-obvious for a user to know to come to this repo and examine the docs folder.

This PR assumes that `/docs/web-configuration.md` is intended to be a stable URL and directs users to it so they know how to populate the file.

Tagging @SuperQ and @roidelapluie per `CONTRIBUTING.md`


SuperQ (approved):
Thanks!

SuperQ:
I wonder if we should integrate this into the prometheus.io site somewhere to shorten up the URL. Either way, this is fine for now.

PR #147, opened 2023-04-01
Add toolkit package

SuperQ:
Implement a top level package to make it easy to bootstrap an exporter.
* Move the flag package to the top level.
* Add support for `--web.telemetry-path` flag. Defaults to`/metrics`.
* Add a self-check function to the web FlagConfig.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -10,7 +10,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-package kingpinflag
+package toolkit
 
 import (
 	"runtime"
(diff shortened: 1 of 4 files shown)


SuperQ:
Here's what a basic startup looks like:

```go
	landingConfig := web.LandingConfig{
		Name:        "Node Exporter",
		Description: "Prometheus Node Exporter",
		Version:     version.Info(),
	}
	toolkitConfig := toolkit.Config{
		MetricsHandler: newHandler(!*disableExporterMetrics, *maxRequests, logger),
		FlagConfig:     toolkitFlags,
		LandingConfig:  landingConfig,
		Logger:         logger,
	}
	if err := toolkit.Run(toolkitConfig); err != nil {
		os.Exit(1)
	}
```

SuperQ:
Still some work to do, tests, etc. But what do you think of this concept so far?

SuperQ:
Ok, new example setup. There's an annoying chicken-and-egg problem with the MetricsHandler, since if we want to pass a logger to the handler, we need to `New()` first.

```
  toolkitConfig := toolkit.Config{
    Name:           "Node Exporter",
    Description:    "Prometheus Node Exporter",
    DefaultAddress: ":9100",
  }
  t := toolkit.New(toolkitConfig)
  t.SetMetricsHandler(newHandler(!*disableExporterMetrics, *maxRequests, t.Logger))

  if err := t.Run(); err != nil {
    level.Error(t.Logger).Log("err", err)
    os.Exit(1)
  }
```

SuperQ:
One solution, I could implement part of the node_exporter's handler into this. One of the nice features is that it would implement the `--web.disable-exporter-metrics` flag by default, which would make a lot of users happy.

ArthurSens:
Hey hey, closing since #394 got merged :)

Issue #157, opened 2023-06-06
Support TLS settings to be passed inline


fayzal-g:
Recently, a change was accepted into `prometheus/common` to allow TLS settings to be passed in-line: see (https://github.com/prometheus/common/pull/472)

It'd be great if it would be possible to also have something similar in this library. Therefore extending the `TLSConfig` to something like:

```
type TLSConfig struct {
        TLSCert string `yaml:"cert"`        
        TLSKey string `yaml:"key"`
        ClientCAText string `yaml:"client_ca"`

	TLSCertPath              string     `yaml:"cert_file"`
	TLSKeyPath               string     `yaml:"key_file"`
	ClientAuth               string     `yaml:"client_auth_type"`
	ClientCAs                string     `yaml:"client_ca_file"`
	CipherSuites             []Cipher   `yaml:"cipher_suites"`
	CurvePreferences         []Curve    `yaml:"curve_preferences"`
	MinVersion               TLSVersion `yaml:"min_version"`
	MaxVersion               TLSVersion `yaml:"max_version"`
	PreferServerCipherSuites bool       `yaml:"prefer_server_cipher_suites"`
	ClientAllowedSans        []string   `yaml:"client_allowed_sans"`
}
```

and the subsequent changes to ensure there are no conflicts between files and inline values. The inline settings would be expected to be the same as the file contents, a PEM-encoded sequence of bytes as a string. 

If there are no major concerns, I'd be happy to work on this. 

Thanks!

bboreham:
What is left to be done here after #158 ?

PR #158, opened 2023-06-09
Allow passing of TLS certificates inline

fayzal-g:
Was suggested in https://github.com/prometheus/exporter-toolkit/issues/157

This PR allows the `TLSCert`, `TLSKey` and `ClientCAs` to be passed inline as a string as part of the `TLSConfig`. The contents of these is expected to be exactly the same as a file, a PEM-encoded sequence of bytes.


SuperQ (changes requested):
Please update `/docs/web-configuration.md`

SuperQ (approved):
Thanks!

roidelapluie:
Thanks!

PR #164, opened 2023-07-19
Remove experimental in CLI

lucacome:
Removes experimental from the CLI flag, as it has been there for 3 years now.


SuperQ (approved):
Agreed, I think the file format is stable enough.

PR #166, opened 2023-07-23
Update to yaml.v3

mikelolasagasti:
Also, modify `empty_config_yml` because with yaml.v3 reading empty file causes Decode to return EOF error, as described in https://github.com/go-yaml/yaml/issues/805 issue.


roidelapluie:
What is the upside of moving to V3?

mikelolasagasti:
Better, somehow, maintained and homogeneity with other prometheus projects.

lucacome:
can we get some traction on this one?

SuperQ:
We've switched to `go.yaml.in/yaml`.

PR #177, opened 2023-11-17
Rename web-configuration.md to web-configuration.netsa0

Netsa-1989:


metalmatze:
:thinking: 
The file is a markdown file. I don't see how this change makes sense. How would this improve things?

PR #183, opened 2023-12-05
Update CI

SuperQ:
* Split golangci-lint to separate action file so it's auto-updated.
* Update Go version in CI.
* Update minimum Go version.

### DIFF
--- a/web/cache.go
+++ b/web/cache.go
@@ -18,14 +18,9 @@ package web
 import (
 	weakrand "math/rand"
 	"sync"
-	"time"
 )
 
 var cacheSize = 100
-
-func init() {
-	weakrand.Seed(time.Now().UnixNano())
-}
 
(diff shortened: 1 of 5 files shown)


roidelapluie (approved):

web/cache.go
```diff
@@ -18,15 +18,10 @@ package web
 import (
 	weakrand "math/rand"
 	"sync"
-	"time"
 )
 
 var cacheSize = 100
 
-func init() {
-	weakrand.Seed(time.Now().UnixNano())
```
roidelapluie, line 27: is this not needed anymore?

SuperQ: No, deprecated and not needed since Go 1.20.

SuperQ:
I'm not sure what's up with the 1.21 failure.

PR #190, opened 2023-12-19
Add hints

lucacome:
Adds hints for completion.


web/kingpinflag/flag.go
```diff
@@ -28,13 +28,13 @@ func AddFlags(a *kingpin.Application, defaultAddress string) *web.FlagConfig {
 		systemdSocket = a.Flag(
 			"web.systemd-socket",
 			"Use systemd socket activation listeners instead of port listeners (Linux only).",
-		).Bool()
+		).HintOptions("true", "false").Bool()
```
SuperQ, line 31: How does this work for kingpin Bool flags? They have a unique style of `--[no-]web.systemd-socket`. Not `--web.systemd-socket=true`.

lucacome: I removed it, it doesn't work for bool flags

SuperQ (approved):
Nice

PR #193, opened 2024-01-07
feat: HTTP Bearer Authorization for simple use cases

robbat2:
Add support for HTTP Bearer Authorization for simple use cases, where HTTP Basic might not fit workflows.

Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>


SuperQ:
I wonder if we should support arbitrary headers, rather than specifically one kind of bearer token. The reason we have the basic auth here is because there's some standard encoding involved. `Authorization: Bearer xxxx` is a simple opaque string.

robbat2:
I actually started considering other auth to add, including HTTP Digest (which also needs special handling like Basic)

For your proposal, it would be match header to accept:
```
match_header:
  authorization:
    - "Bearer foo"
    - "FakeAuth value2"
  x-random-stuff:
    - "abc123"
```

SuperQ:
Some work on Digest auth was being done over in the common package.

https://github.com/prometheus/common/pull/553

robbat2:
@SuperQ  hmm, reading that, I realize there is more auth support in common already that not being exposed in toolkit; should I include that in this PR?

roidelapluie:
> @SuperQ hmm, reading that, I realize there is more auth support in common already that not being exposed in toolkit; should I include that in this PR?

Anything beyond basic auth would require you to set a proxy. The exporter toolkit is limited to: TLS, and simple use case. The idea is to enable you to have a trusted connection between your exporter and the proxy of your choice.

PR #196, opened 2024-01-16
Add pprof links to landing page

SuperQ:
Make it easier for users to fetch and download pprof data for debugging.
* Add a link to https://pprof.me for easy sharing.

image

### DIFF
--- a/web/landing_page.html
+++ b/web/landing_page.html
@@ -30,6 +30,14 @@
       </div>
       {{ end }}
       {{ .ExtraHTML }}
+      <div id="pprof">
+      Download a detailed report of resource usage (pprof format, from the Go runtime):
+      <ul>
+        <li><a href="debug/pprof/heap">heap usage (memory)</a>
+        <li><a href="debug/pprof/profile?seconds=60">CPU usage (60 second profile)</a>
+      </ul>
+      To visualize and share profiles you can upload to <a href="https://pprof.me" target="_blank">pprof.me</a>
+      </div>
     </main>
   </body>
 </html>
--- a/web/landing_page.css
+++ b/web/landing_page.css
@@ -15,4 +15,9 @@ label {
   display: inline-block;
   width: {{.Form.Width}}em;
 }
+#pprof {
+  border: black 2px solid;
+  padding: 1rem;
+  width: fit-content;
+}
 {{.ExtraCSS}}


metalmatze (approved):
That is an excellent idea!
It will make it a lot easier to get profiles from users.

SuperQ:
@bboreham What do you think of adding something like this to Prometheus `/status`?

web/landing_page.html
```diff
@@ -30,6 +30,14 @@ <h1>{{.Name}}</h1>
       </div>
       {{ end }}
       {{ .ExtraHTML }}
+      <div id="pprof">
+      Go pprof debugging profile data:
+      <ul>
+        <li><a href="debug/pprof/heap">heap usage (memory)</a>
+        <li><a href="debug/pprof/profile">CPU usage (takes some time to gather)</a>
+      </ul>
+      Try posting these to <a href="https://pprof.me" target="_blank">pprof.me</a>
```
bboreham, line 39:
```suggestion
       "To share with an expert, you can upload to <a href="https://pprof.me" target="_blank">pprof.me</a>
```

web/landing_page.html
```diff
@@ -30,6 +30,14 @@ <h1>{{.Name}}</h1>
       </div>
       {{ end }}
       {{ .ExtraHTML }}
+      <div id="pprof">
+      Go pprof debugging profile data:
```
bboreham, line 34:
```suggestion
      To download a detailed report of resource usage (pprof format, from the Go runtime):
```

web/landing_page.html
```diff
@@ -30,6 +30,14 @@ <h1>{{.Name}}</h1>
...
+      <ul>
+        <li><a href="debug/pprof/heap">heap usage (memory)</a>
+        <li><a href="debug/pprof/profile">CPU usage (takes some time to gather)</a>
```
bboreham, line 37: You could add `?seconds=10` to make it less alarming for the newcomer.

bboreham (commented):
Good idea; I had some suggestions about wording.

What would you think if I said it should offer both pprof.me and flamegraph.com ?

bboreham:
> What do you think of adding something like this to Prometheus `/status`?

Sure.  I guess the page with most info about memory is `/tsdb-status`.  Suspect what people really want is a top-level breakdown like "40% on samples, 30% on scraping, 20% on remote-write".  Which we don't know how to do accurately, but might be able to approximate in a useful way.

web/landing_page.html
SuperQ: Hmm, I don't like the `expert` term. I made a minor tweak to the suggestion.

web/landing_page.html
SuperQ: Yes, but that has a high chance of missing a scrape interval. I was actually thinking about making this `60s` so we gather a few scrapes. Or capture instances where people reduce the scrape interval.

dswarbrick:
It might be worth mentioning somewhere in docs that these links will be ineffectual (and most likely just bounce back to the landing page, albeit with additional path suffixes each time) unless the exporter imports `net/http/pprof`.

SuperQ:
My plan is to add pprof handling to the exporter-toolkit.

PR #202, opened 2024-02-26
Support listen on vsock

jeffrey4l:
metrics can be retrive without nic between host and qemu guest through vsock.

more info check

* https://man7.org/linux/man-pages/man7/vsock.7.html
* https://wiki.qemu.org/Features/VirtioVsock

### DIFF
--- a/web/kingpinflag/flag.go
+++ b/web/kingpinflag/flag.go
@@ -33,7 +33,7 @@ func AddFlags(a *kingpin.Application, defaultAddress string) *web.FlagConfig {
 	flags := web.FlagConfig{
 		WebListenAddresses: a.Flag(
 			"web.listen-address",
-			"Addresses on which to expose metrics and web interface. Repeatable for multiple addresses.",
+			"Addresses on which to expose metrics and web interface. Repeatable for multiple addresses. Examples: `:9100` or `127.0.0.1:9100` for http, `vsock://:9100` for vsock",
 		).Default(defaultAddress).HintOptions(defaultAddress).Strings(),
 		WebSystemdSocket: systemdSocket,
 		WebConfigFile: a.Flag(
--- a/web/tls_config.go
+++ b/web/tls_config.go
@@ -20,12 +20,16 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"github.com/coreos/go-systemd/v22/activation"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/mdlayher/vsock"
 	config_util "github.com/prometheus/common/config"
 	"golang.org/x/sync/errgroup"
 	"gopkg.in/yaml.v2"
@@ -275,9 +279,11 @@ func ServeMultiple(listeners []net.Listener, server *http.Server, flags *FlagCon
 }
 
 // ListenAndServe starts the server on addresses given in WebListenAddresses in
-// the FlagConfig or instead uses systemd socket activated listeners if
-// WebSystemdSocket in the FlagConfig is true. The FlagConfig is also passed on
-// to ServeMultiple.
+// the FlagConfig. When address starts looks like vsock://:{port}, it listens on
+// vsock. More info check https://wiki.qemu.org/Features/VirtioVsock .
+// Or instead uses systemd socket activated listeners if WebSystemdSocket in the
+// FlagConfig is true.
+// The FlagConfig is also passed on to ServeMultiple.
 func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) error {
 	if flags.WebSystemdSocket == nil && (flags.WebListenAddresses == nil || len(*flags.WebListenAddresses) == 0) {
 		return ErrNoListeners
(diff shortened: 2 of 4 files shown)


web/tls_config.go
```diff
@@ -297,7 +301,17 @@ func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) e
 
 	listeners := make([]net.Listener, 0, len(*flags.WebListenAddresses))
 	for _, address := range *flags.WebListenAddresses {
-		listener, err := net.Listen("tcp", address)
+		var err error
+		var listener net.Listener
+		if strings.HasPrefix(address, "vsock://") {
+			var port uint32
+			port, err = parseVsockAddress(address)
```
SuperQ, line 308:
```suggestion
			port, err := parseVsockAddress(address)
```

SuperQ (changes requested):
Interesting idea. This also needs to be added to the Go documentation and probably the flag help string in `web/kingpinflag/flag.go`.

web/tls_config.go
SuperQ: I'm pretty sure it should be ok, it looks like you applied the change.

jeffrey4l: technically , i move the `if err != nil { return err}` into the if scope ( which can use := ) 
but listener still require `=` to make sure it can be used out of if-else scope.

Anyway, current code works as expected.

SuperQ:
Please add documentation to the `ListenAndServe` func docs (before like 285). A link to https://wiki.qemu.org/Features/VirtioVsock would be useful there.

web/tls_config.go
SuperQ: Yes, `listener` still needs a declared var.

SuperQ (approved):
Thanks, looks good.

bwplotka:
👍🏽

SuperQ:
Would you please rebase to fix the conflict?

jesusvazquez (approved):
LGTM

I'd like some testing for ListenAndServe here https://github.com/prometheus/exporter-toolkit/blob/48431d7bea401fa242dfa94b9f36c0efa039dd69/web/tls_config_test.go#L479-L492

To make sure it listens both on tcp and vsock if specified.

But I won't block on this if other people don't think this is necessary 👍

web/tls_config.go
```diff
@@ -297,16 +303,45 @@ func ListenAndServe(server *http.Server, flags *FlagConfig, logger log.Logger) e
...
+				return err
+			}
+			listener, err = vsock.Listen(port, nil)
+			if err != nil {
+				return err
+			}
+		} else {
+			listener, err = net.Listen("tcp", address)
+			if err != nil {
+				return nil
```
Leonid99, line 320: Shouldn't this be `return err`? It silently swallows an error condition.

web/kingpinflag/flag.go
```diff
@@ -33,7 +33,7 @@ func AddFlags(a *kingpin.Application, defaultAddress string) *web.FlagConfig {
 	flags := web.FlagConfig{
 		WebListenAddresses: a.Flag(
 			"web.listen-address",
-			"Addresses on which to expose metrics and web interface. Repeatable for multiple addresses.",
+			"Addresses on which to expose metrics and web interface. Repeatable for multiple addresses. Examples: `:9100` or `127.0.0.1:9100` for http, `vsock://:9100` for vsock",
```
Leonid99, line 36: I think, vsock having an URI-like format is confusing and asymmetric, especially given that it doesn't really support the host portion of the URI. When a user sees that listening on a vsock expects `vsock://`, they would assume that listening on a TCP port expects `--web.listen-address=http://:9100`, but that's not the case.
Moreover, this help message is confusing, because the vsock connection also uses http. In terms of network layers, vsock is most similar to IP.

I would recommend to change the expected format to `--web.listen-address=vsock:9100`, to avoid confusion.

SuperQ: I don't think I like the `vsock:9100` format. I much prefer the URI format.

septatrix (commented):
This behavior can already be archived using the existing systemd socket use case as those are not only able to listen on TCP/IP (AF_INET/AF_INET6) and Unix Domain Sockets (AF_UNIX), but also on Vsock (AF_VSOCK) - among some others.

SuperQ:
@septatrix 

Interesting, so you're saying that this isn't needed? Do we just need to update the documentation?

septatrix:
> @septatrix
> 
> Interesting, so you're saying that this isn't needed? Do we just need to update the documentation?

Yes, at least for systems using systemd (or a compatible implementation of the listener interface). See [`man:systemd.socket`](https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html#ListenStream=):

> If the address string is a string in the format "vsock:x:y", it is read as CID x on a port y address in the AF_VSOCK family. The CID is a unique 32-bit integer identifier in AF_VSOCK analogous to an IP address. Specifying the CID is optional, and may be set to the empty string. "vsock" may be replaced with "vsock-stream", "vsock-dgram" or "vsock-seqpacket" to force usage of the corresponding socket type.

Leonid99:
> that this isn't needed

I would advocate to merge this change anyway. Using systemd for sockets translation isn't different from other methods of socket translation, such as `socat`, i.e. it's always possible, for example, expose a socket that listens on a TCP port to local domain sockets, etc., but most network services support listening on both natively, without `socat` or systemd.

Also, another reminder that a rebased version of this PR exists as #220. (I am not trying to claim credit, I'm just interested in having this merged).

SuperQ:
Closing in favor of #220

PR #206, opened 2024-03-01
Update Go

SuperQ:
* Update Go to 1.22.
* Update Go modules.
* Update tests for cipher suites removed in Go 1.22.


SuperQ:
@roidelapluie if you have a chance to look. It seems like Go has disabled some more TLS options. Should I just remove the broken tests?

go.mod
```diff
@@ -1,13 +1,13 @@
...
-go 1.20
+go 1.21
 
 require (
 	github.com/alecthomas/kingpin/v2 v2.4.0
 	github.com/coreos/go-systemd/v22 v22.5.0
 	github.com/go-kit/log v0.2.1
-	github.com/prometheus/common v0.46.0
-	golang.org/x/crypto v0.20.0
+	github.com/prometheus/common v0.50.0
```
ArthurSens, line 9: I'm not too familiar with all that exporter toolkit does, does it add any metrics to your application? If yes, we probably want to avoid common 0.50 🤔

SuperQ: Oh, oops, I made this weeks ago. :joy_cat:

ArthurSens (approved):
I'm not familiar with the TLS changes, but giving the approval in case you just want the rubber stamp :)

PR #226, opened 2024-06-28
Update go.mod and deps

lucacome:
Since version `1.21` go requires specifying the patch in go.mod https://go.dev/doc/toolchain#version

This also bumps the dependencies, I think dependabot was stuck because of the above, but can't confirm since I don't have access to it.


SuperQ:
I think you are confusing the `go` directive and `toolchain` directive.

https://go.dev/ref/mod#go-mod-file-go

We use dependabot to manage updates.

lucacome:
@SuperQ I'm not confusing it, see also https://github.com/golang/go/issues/62278#issuecomment-1693538776

You can even test this yourself if you do `go mod init` on a new project with go 1.21+ you will get the patch version. 
Or in this one for example, remove `go.mod` and then do 

```
❯ go mod init github.com/prometheus/exporter-toolkit
go: creating new go.mod: module github.com/prometheus/exporter-toolkit
go: to add module requirements and sums:
	go mod tidy
```
will result in 
```
❯ cat go.mod
module github.com/prometheus/exporter-toolkit

go 1.22.5
```

PR #227, opened 2024-06-28
Add dependabot for Actions

lucacome:
Adds GitHub Actions to the list


lucacome:
I see that `golangci-lint.yml` is synced from `prometheus/prometheus` but there's no way to exclude a single file AFAIK

SuperQ:
Yes, we currently sync that file, so this won't be possible.

PR #245, opened 2024-09-01
Update build

SuperQ:
* Update Go minimum to 1.22.
* Enable dependabot for github actions.
* Bump promci action.


lucacome:
I'm just wondering what changed since https://github.com/prometheus/exporter-toolkit/pull/227

SuperQ:
We now have a ci.yml that needs to be maintained. Sadly dependabot can't exclude single files so the synced file will get update PRs which we have to ignore. :(

PR #258, opened 2024-10-02
basic auth: add option to exclude bcrypt at build time

jan--f:
This add the option to exclude bcrypt during the build process by passing `-tags nobcrypt`. Currently this disables user authentication via basic_auth. If authorized users are configured but the server is build with `nobcrypt` the endpoint will respond with Unauthorized, even if the correct credentials are send.
The goal is to be able to exclude bcrypt for compliance reasons. In the future we could add basic_auth using other hash functions.

### DIFF
--- a/web/basic_auth/no_bcrypt.go
+++ b/web/basic_auth/no_bcrypt.go
@@ -1,0 +1,35 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build nobcrypt
+// +build nobcrypt
+
+package basic_auth
+
+import (
+	"fmt"
+	"log/slog"
+
+	config_util "github.com/prometheus/common/config"
+)
+
+func Validate(users map[string]config_util.Secret) error {
+	if len(users) > 0 {
+		slog.Info("basic auth via bcrypt hashes not implemented")
+	}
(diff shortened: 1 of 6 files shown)


web/basic_auth/no_bcrypt.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+	"log/slog"
+
+	config_util "github.com/prometheus/common/config"
+)
+
+func Validate(users map[string]config_util.Secret) error {
+	if len(users) > 0 {
+		slog.Info("basic auth via bcrypt hashes not implemented")
```
jan--f, line 28: This cloud also return an error so a server without bcrypt support can not start with a config that includes authorized users.

matthiasr: I think that would be better, if we know at startup time that a configuration cannot work, failing explicitly lets the operator figure out what they want to do about it.

matthiasr:
Just as a note, this will have merge conflicts with #151 which also refactors how authentication happens. Can we get that one in before finishing up here?

matthiasr (commented):
Now I am curious what compliance regime _forbids_ bcrypt 😄 and at least this is not as difficult as FIPS mode would be …

jan--f:
> Just as a note, this will have merge conflicts with #151 which also refactors how authentication happens. Can we get that one in before finishing up here?

Sure no problem, I'm not in a rush. This might not be the best way anyway.

jan--f:
> Now I am curious what compliance regime _forbids_ bcrypt 😄 and at least this is not as difficult as FIPS mode would be …

This does actually target FIPS. Afaiu `bcrypt` is at least not in the list of approved crypto algos. But most certain depending on `x/crypto` is a problem. :shrug:

matthiasr:
Hmm, yeah, TLS won't be an option then either. I guess in our case FIPS-mode would entail no encryption at all 😅 I think #151 will actually help in this because it adds the notion of chained authenticators, which adds a well contained point to control what is and isn't active, and then we can remove the relevant authenticators from the build altogether.

jan--f:
@matthiasr afaiu using golangs `crypt` module is fine and can be fips compliant, at least starting with go1.24. See also https://go.dev/doc/security/fips140

PR #282, opened 2025-01-07
Feat: add possibility to disable pprof form

odnobit:
This pull request introduces a feature to disable the `pprof` form on the landing page.

The new `DisablePprofForm` boolean parameter defaults to false, ensuring backward compatibility with existing behavior.

### DIFF
--- a/web/landing_page.go
+++ b/web/landing_page.go
@@ -27,15 +27,16 @@ import (
 
 // Config represents the configuration of the web listener.
 type LandingConfig struct {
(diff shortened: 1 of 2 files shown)


odnobit:
@SuperQ @roidelapluie WDYT?

SuperQ:
Why? This doesn't seem to solve any issue.

odnobit:
If pprof is not enabled, the links will return a 404 error, which I find problematic

SuperQ:
I don't see this as necessary.

PR #283, opened 2025-01-10
Serve behind route prefix

kwilt:
This new parameter will allow for serving the landing page when using --web.external-url as an argument with many exporters. 

I added this as it is needed for this pull request: https://github.com/prometheus/snmp_exporter/pull/1335

So in practice it will look something like this:

```go
landingConfig := web.LandingConfig{
        // config stuff goes here ...
		}
                                                                      // new parameter
		landingPage, err := web.NewLandingPage(landingConfig, *routePrefix != "")
		if err != nil {
			logger.Error("Error creating landing page", "err", err)
			os.Exit(1)
		}
		http.Handle(*routePrefix, landingPage)
```

### DIFF
--- a/web/landing_page.go
+++ b/web/landing_page.go
@@ -61,7 +61,8 @@ type LandingLinks struct {
 }
 
 type LandingPageHandler struct {
-	landingPage []byte
+	landingPage   []byte
+	routePrefixed bool
 }
 
 var (
@@ -71,7 +72,7 @@ var (
 	landingPagecssContent string
 )
 
-func NewLandingPage(c LandingConfig) (*LandingPageHandler, error) {
+func NewLandingPage(c LandingConfig, routePrefixed bool) (*LandingPageHandler, error) {
 	var buf bytes.Buffer
 
 	length := 0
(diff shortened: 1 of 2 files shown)


web/landing_page.go
```diff
@@ -71,7 +72,7 @@ var (
 	landingPagecssContent string
 )
 
-func NewLandingPage(c LandingConfig) (*LandingPageHandler, error) {
+func NewLandingPage(c LandingConfig, routePrefixed bool) (*LandingPageHandler, error) {
```
SuperQ, line 75: We already have a `LandingConfig`, it doesn't make sense to add another `bool` value here.

Also, rather than have a bool value, we can add a `RoutePrefix string` directly to the `LandingConfig`. If the `RoutePrefix` is `""`, we can default it to `/`.

SuperQ (changes requested):

kwilt:
Okay, I think I've made the changes you requested

web/landing_page.go
```diff
@@ -106,10 +115,6 @@ func NewLandingPage(c LandingConfig) (*LandingPageHandler, error) {
 }
 
 func (h *LandingPageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Path != "/" {
```
SuperQ, line 109: We need to keep some form of this otherwise the landing page handler will respond to `/routeprefix/somerandompage` with content. We want to 404 anything that doesn't match the RoutePrefix.

SuperQ:
Looking good so far, thanks!

web/landing_page.go
```diff
@@ -93,6 +96,13 @@ func NewLandingPage(c LandingConfig) (*LandingPageHandler, error) {
 		}
 		c.CSS = buf.String()
 	}
+	if c.RoutePrefix == "" {
+		c.RoutePrefix = "/"
+	}
+	// Strip leading '/' from Links if present
```
SuperQ, line 102: Hmm, should we also enforce that RoutePrefix has a trailing `/`?

kwilt: Probably a good idea

kwilt:
Anything else you'd like fixed/added to this? 😄

SuperQ (approved):
Thanks!

PR #286, opened 2025-01-21
Allow `web.listen-address` to be unix domain sockets

nkcmr:
## Context / Background
Prometheus push gateway is used extensively in the systems I work on. The current setup I use is to have the pushgateway container running adjacent to my application and to push metrics over the TCP loopback interface. As the usage of the pushgateway has scaled up, the overhead and reliability have become worse.

Using unix domain sockets would allow metrics to be collected in a much quicker and reliable way.

 ## In this PR
- Added an ability to set `--web.listen-address` to `unix:/path/to/sock` as a way of specifying that the server should open a listener using "unix" domain sockets.

 ## How this was tested
- I have made patches against prometheus push gateway and made a proof-of-concept that seems to work correctly.

### DIFF
--- a/web/kingpinflag/flag.go
+++ b/web/kingpinflag/flag.go
@@ -39,7 +39,7 @@ func AddFlags(a flagGroup, defaultAddress string) *web.FlagConfig {
 	flags := web.FlagConfig{
 		WebListenAddresses: a.Flag(
 			"web.listen-address",
-			"Addresses on which to expose metrics and web interface. Repeatable for multiple addresses. Examples: `:9100` or `[::1]:9100` for http, `vsock://:9100` for vsock",
+			"Addresses on which to expose metrics and web interface. Repeatable for multiple addresses. Examples: `:9100` or `[::1]:9100` for http, `vsock://:9100` for vsock, `unix:/var/run/prom.sock` for unix domain sockets",
 		).Default(defaultAddress).HintOptions(defaultAddress).Strings(),
 		WebSystemdSocket: systemdSocket,
 		WebConfigFile: a.Flag(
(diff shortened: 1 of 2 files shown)


nkcmr:
@SuperQ Hello! 👋 — I've seen you be the one to merge stuff around here. Mind if I get a review/feedback whenever you get a chance?

SuperQ:
That seems like an [XY Problem](https://xyproblem.info/). If pushing to the pushgateway is a performance bottleneck, perhaps you should not be using the pushgateway in the first place.

The pushgateway is _not_ meant as a high performance metrics sink.

nkcmr:
@SuperQ Thanks for the feedback.

---

> The pushgateway is not meant as a high performance metrics sink.

There is not much wrong with that logic.

If you were a co-worker I might be able to make an appeal to reason about how this system I was describing had a lot of "legacy-ness" about it and refactoring the metrics collection was a lot of work and how this would just buy some time to delay that work. But that isn't the case so that is not an effective argument.

The only remaining argument I have then is that, there are probably a lot of other reasons that someone might want to listen on a unix domain socket that are valid. Perhaps they want a metrics server to be listening, but all of its traffic handled through some other proxy (NGINX) and for security reasons want the upstream connection to be a socket?

My overarching point is that this is a fairly trivial and small change that for a low cost can open other use cases for someone else in the future and definitely me, right now.

I understand that is quite a weak argument, so if "no" is still your answer, then I am totally fine with that and I'll just be on my way and go build out some way to make my own patch setup and achieve my use-case anyways.

Sorry for the rambling response and thanks again for taking the time to review.

PR #287, opened 2025-01-21
Add security attributes to external links to prevent reverse tabnabbing

erlan-z:
Adding rel="noopener noreferrer" to external links that use target="_blank" to prevent potential reverse tabnabbing attacks. This security enhancement blocks malicious websites opened in new tabs from manipulating the original page context.


SuperQ (approved):
Thanks!

PR #293, opened 2025-02-06
Add processing for route prefixes, external urls, and pprof mux

kwilt:
Supporting pull request for https://github.com/prometheus/snmp_exporter/pull/1335

Maybe this sucks, idk. I'm getting out of my comfort zone with this pull request lol. But I gave it a shot anyway. 😬 

This is my attempt at moving a good chunk of the routing for the exporter into the landing page's codebase. In theory this could make a more cohesive experience across exporters and allow for easier development as most of the routing work is now done by using the LandingConfig struct.


ArthurSens (commented):
hey, sorry for the super delayed review here. Looks like the codebase has changed quite a bit since this PR was opened.

Route prefix was already added in #283, right? So this PR conflicts with your old one.

It is also adding unconditional addition of pprof endpoint, which I honesly like the idea, but we're doing way too many separate changes at once. Let's keep things separate.

Following the current codebase, the `bootstrap` package now owns the the http mux, let's make changes over there :)

With all that said, I'll take the liberty to close this PR. I dont intend to discourage you of making the changes here, it's just because the PR is quite old and starting from scratch will be faster than rebasing the current work!

PR #327, opened 2025-06-17
remove gopkg.in/yaml.v2

miparnisari:


mrueg:
Maybe it's better to replace it with https://github.com/yaml/go-yaml

PR #387, opened 2026-03-28
SPIFFE support

### LINKED ISSUE #259: SPIFFE support
The SPIFFE Workload api can be used to fetch and keep up to date tls certs/keys (Default 1 hour validity window). I would like to be able to have a flag to just enable SPIFFE support in the TLSConfig and optionally give it a list of spiffe id's to allow access.

kfox1111:
This feature allows the service to be MTLS secured automatically using SPIFFE/SPIRE.

```
tls_server_config:
  spiffe:
    enabled: true
    socket_path: "unix:///tmp/spire-agent/public/api.sock"

    # optional list of spiffeids to allow.
    # if not specified, spiffe://<trust_domain>/prometheus> will be validated
    #authorized_ids:
    # -  spiffe://example.org/prometheus1
    # -  spiffe://example.org/prometheus2
```

fixes: https://github.com/prometheus/exporter-toolkit/issues/259


SuperQ:
This is interesting, but pulls in gRPC, which pulls in a bunch of dependencies. Not sure about this.

kfox1111:
@SuperQ what if we put it behind a disable_spiffe build tag? Then those that don't want the dependency can compile it out and those that don't care, gain the extra functionality out of the box?

SuperQ:
The main issue is I think we would want the consistency of having it always available.

kfox1111:
I would love for it to be always available, as an end user. The ability to flip a switch and security is taken care of is awesome.

PR #394, opened 2026-05-21
feat: exporter toolkit bootstrap

nicolastakashi:
## Summary
This adds a new optional bootstrap package to exporter-toolkit for exporters that want a shared startup path for common process wiring. Today, many exporters duplicate the same startup concerns around:
- common web flags
- logger setup
- metrics path registration
- landing page wiring
- web.ListenAndServe integration

This PR introduces an additive package that can own those shared pieces without changing the existing web package API or forcing any current exporter to migrate.

## Why
The goal is to reduce repeated startup code across exporters while keeping boundaries clear:
- web remains focused on transport/server concerns such as listeners, TLS, auth, and serving
- bootstrap owns higher-level exporter startup concerns such as flag registration, logging initialization, metrics endpoint wiring, and landing page setup

This is intentionally additive:
- exporters already using web.ListenAndServe can continue unchanged
- exporters can opt into bootstrap only if it simplifies their startup path
- this gives us a concrete shared layer for downstream adoption without requiring ecosystem-wide churn

A key design choice in this PR is keeping web.telemetry-path ownership in bootstrap, not in web.FlagConfig, so transport concerns and exporter routing concerns do not get mixed together. It also keeps the existing web startup path intact.

### Startup Example

**Without bootstrap:**

```go
metricsPath := kingpin.Flag(
	"web.telemetry-path",
	"Path under which to expose metrics.",
).Default("/metrics").String()
disableExporterMetrics := kingpin.Flag(
	"web.disable-exporter-metrics",
	"Exclude metrics about the exporter itself (promhttp_*, process_*, go_*).",
).Bool()
maxRequests := kingpin.Flag(
	"web.max-requests",
	"Maximum number of parallel scrape requests. Use 0 to disable.",
).Default("40").Int()
toolkitFlags := kingpinflag.AddFlags(kingpin.CommandLine, ":9100")
promslogConfig := &promslog.Config{}
flag.AddFlags(kingpin.CommandLine, promslogConfig)
kingpin.Parse()
logger := promslog.New(promslogConfig)

http.Handle(*metricsPath, newHandler(!*disableExporterMetrics, *maxRequests, logger))
if *metricsPath != "/" {
	landingPage, err := web.NewLandingPage(web.LandingConfig{
		Name:        "My Exporter",
		Description: "Prometheus My Exporter",
		Version:     version.Info(),
		Links: []web.LandingLinks{{
			Address: *metricsPath,
			Text:    "Metrics",
		}},
	})
	if err != nil {
		os.Exit(1)
	}
	http.Handle("/", landingPage)
}

server := &http.Server{}
if err := web.ListenAndServe(server, toolkitFlags, logger); err != nil {
	os.Exit(1)
}
```

**With bootstrap:**

```go
runner := bootstrap.New(bootstrap.Config{
	App:            kingpin.CommandLine,
	Name:           "my_exporter",
	Description:    "Prometheus My Exporter",
	DefaultAddress: ":9100",
	MetricsHandlerFactory: func(b *bootstrap.Bootstrap) (http.Handler, error) {
		return newHandler(!b.DisableExporterMetrics, b.MaxRequests, b.Logger), nil
	},
})
if err := runner.Run(); err != nil {
	fmt.Fprintln(os.Stderr, err)
	os.Exit(1)
}
```
Notes
- This PR does not deprecate or remove the existing web startup flow.
- This PR does not try to model exporter-specific runtime config.
- This PR is intended to be proven by real downstream adoption, starting with node_exporter.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1,0 +1,259 @@
+// Copyright 2026 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package bootstrap provides a small bootstrap layer for exporters that want the
+// common exporter-toolkit web flags, logging setup, landing page wiring, and
+// listener startup behavior.
+
+package bootstrap
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"os"
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/prometheus/common/promslog"
+	promslogflag "github.com/prometheus/common/promslog/flag"
+	"github.com/prometheus/common/version"
+	"github.com/prometheus/exporter-toolkit/web"
+	"github.com/prometheus/exporter-toolkit/web/kingpinflag"
+)
+
+var (
+	// ErrNoMetricsHandler is returned when no metrics handler source was configured.
+	ErrNoMetricsHandler = errors.New("missing metrics handler")
+	// ErrMultipleMetricsSource is returned when both a static handler and a
+	// handler factory were configured.
+	ErrMultipleMetricsSource = errors.New("only one metrics handler source may be configured")
+	// ErrMissingMetricsPathFlag is returned when the metrics path flag was not initialized.
+	ErrMissingMetricsPathFlag = errors.New("missing metrics path flag configuration")
+)
+
+// MetricsHandlerFactory builds an exporter-specific metrics handler after the
+// common toolkit flags have been parsed.
+type MetricsHandlerFactory func(*Bootstrap) (http.Handler, error)
+
+// Bootstrap captures parsed startup options that handler factories can use to
+// construct exporter-specific HTTP handlers after CLI parsing.
+type Bootstrap struct {
+	// Logger is the configured logger for the exporter process.
+	Logger *slog.Logger
+	// FlagConfig contains the parsed exporter-toolkit web flags.
+	FlagConfig *web.FlagConfig
+	// DisableExporterMetrics reports whether exporter self-metrics should be disabled.
+	DisableExporterMetrics bool
+	// MaxRequests is the parsed value of --web.max-requests.
+	MaxRequests int
+}
+
+// Config defines the generic exporter bootstrap inputs.
+type Config struct {
+	// App is the Kingpin application to register flags on. When nil,
+	// kingpin.CommandLine is used.
+	App *kingpin.Application
+	// Name is the exporter name used for version output and startup logging.
+	Name string
+	// Description is used as the default landing page description.
+	Description string
+	// DefaultAddress is the default value for --web.listen-address.
+	DefaultAddress string
+	// Logger is the logger to use. When nil, toolkit configures promslog flags
+	// and builds a logger during Parse.
+	Logger *slog.Logger
+	// LandingConfig customizes the generated landing page.
+	LandingConfig web.LandingConfig
+	// MetricsHandler is the static handler to register at the metrics path.
+	MetricsHandler http.Handler
+	// MetricsHandlerFactory builds the metrics handler after flags are parsed.
+	MetricsHandlerFactory MetricsHandlerFactory
+}
+
+// Runner manages generic exporter startup around flag parsing, landing page
+// setup, and web listener bootstrapping.
+type Runner struct {
+	app       *kingpin.Application
+	logConfig *promslog.Config
+	provided  Config
+
+	disableExporterMetrics *bool
+	maxRequests            *int
+
+	// Logger is the resolved logger after parsing configuration.
+	Logger *slog.Logger
+	// FlagConfig contains the parsed exporter-toolkit web flags.
+	FlagConfig *web.FlagConfig
+	// DisableExporterMetrics is the parsed value of --web.disable-exporter-metrics.
+	DisableExporterMetrics bool
+	// MaxRequests is the parsed value of --web.max-requests.
+	MaxRequests int
+	// LandingConfig is the resolved landing page configuration.
+	LandingConfig web.LandingConfig
+	// MetricsHandler is the configured static metrics handler.
+	MetricsHandler http.Handler
+	// MetricsHandlerFactory is the configured deferred metrics handler builder.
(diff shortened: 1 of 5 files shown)


SuperQ:
I think this is a continuation of https://github.com/prometheus/exporter-toolkit/pull/147

nicolastakashi:
> I think this is a continuation of #147

Should we get the first one merged first @SuperQ ?

web/tls_config.go
```diff
@@ -66,9 +67,27 @@ type TLSConfig struct {
...
-	WebConfigFile      *string
+	// WebSystemdSocket enables systemd socket activation listeners.
+	WebSystemdSocket *bool
+	// WebConfigFile points to the TLS and authentication configuration file.
+	WebConfigFile *string
+}
+
+// CheckFlags validates that the flag configuration contains the required
+// listener and web config fields needed by the web package.
+func (c *FlagConfig) CheckFlags() error {
```
ArthurSens:
```suggestion
func (c *FlagConfig) checkFlags() error {
```

It doesn't seem to be needed outside of this package

bootstrap/bootstrap.go
```diff
@@ -0,0 +1,270 @@
...
+)
+
+var (
+	// ErrNoMetricsHandler is returned when no metrics handler source was configured.
+	ErrNoMetricsHandler = errors.New("missing metrics handler")
+	// ErrMultipleMetricsSource is returned when both a static handler and a
+	// handler factory were configured.
+	ErrMultipleMetricsSource = errors.New("only one metrics handler source may be configured")
+	// ErrEmptyMetricsPath is returned when the configured metrics path is empty.
+	ErrEmptyMetricsPath = errors.New("metrics path must not be empty")
```
ArthurSens: similar here, do these need to be public?

bootstrap/bootstrap.go
```diff
@@ -0,0 +1,270 @@
...
+	// MetricsHandler is the configured static metrics handler.
+	MetricsHandler http.Handler
+	// MetricsHandlerFactory is the configured deferred metrics handler builder.
+	MetricsHandlerFactory MetricsHandlerFactory
+}
+
+// AddFlags adds the common exporter web flags to a Kingpin application.
+func AddFlags(a *kingpin.Application, defaultAddress string) *web.FlagConfig {
+	return kingpinflag.AddFlags(a, defaultAddress)
+}
```
ArthurSens: Looking at your node exporter PR, you removed `AddFlags` from there, so we might be able to make this private too :)

bootstrap/bootstrap.go
```diff
@@ -0,0 +1,270 @@
...
+	// LandingConfig is the resolved landing page configuration.
+	LandingConfig web.LandingConfig
+	// MetricsHandler is the configured static metrics handler.
+	MetricsHandler http.Handler
+	// MetricsHandlerFactory is the configured deferred metrics handler builder.
+	MetricsHandlerFactory MetricsHandlerFactory
+}
+
+// AddFlags adds the common exporter web flags to a Kingpin application.
+func AddFlags(a *kingpin.Application, defaultAddress string) *web.FlagConfig {
```
ArthurSens:
```suggestion
func AddFlags(a *kingpin.Application, flag string) *web.FlagConfig {
```

looks like this is not only used to register defaultAddress

bootstrap/bootstrap.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	if err := t.FlagConfig.CheckFlags(); err != nil {
+		return err
+	}
+	if *t.metricsPath == "" {
+		return ErrEmptyMetricsPath
+	}
+	if *t.maxRequests < 0 {
+		return fmt.Errorf("web max requests must be greater than or equal to zero")
+	}
```
ArthurSens, line 173: could you clarify why you use a var error at some places and fmt.Errorf in other places?

bootstrap/bootstrap.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil
+}
+
+// Run parses os.Args and starts serving the configured exporter endpoints.
+func (t *Runner) Run() error {
+	return t.RunWithArgs(os.Args[1:])
+}
+
+// RunWithArgs parses the provided args and starts the exporter HTTP server.
+func (t *Runner) RunWithArgs(args []string) error {
```
ArthurSens:
```suggestion
// runWithArgs parses the provided args and starts the exporter HTTP server.
func (t *Runner) runWithArgs(args []string) error {
```

bootstrap/bootstrap.go
```diff
@@ -0,0 +1,270 @@
...
+	if c.Name != "" {
+		app.Version(version.Print(c.Name))
+	}
+	app.HelpFlag.Short('h')
+
+	return t
+}
+
+// Parse parses the provided arguments and resolves the derived bootstrap state.
+func (t *Runner) Parse(args []string) error {
```
ArthurSens:
```suggestion
// parse parses the provided arguments and resolves the derived bootstrap state.
func (t *Runner) parse(args []string) error {
```

ArthurSens (commented):
Really cool! This can be so easily applied to a lot of exporters :)

bootstrap/bootstrap.go
nicolastakashi: I forgot to move those to the var errors.

bootstrap/bootstrap.go
nicolastakashi: Not really.

ArthurSens (approved):
LGTM, but let's see what exporter-toolkit maintainers have to say :)

roidelapluie:
why is  	MetricsHandlerFactory: func(b *bootstrap.Bootstrap) (http.Handler, error) {
		return newHandler(!b.DisableExporterMetrics, b.MaxRequests, b.Logger), nil
	}, needed?

nicolastakashi:
> why is MetricsHandlerFactory: func(b *bootstrap.Bootstrap) (http.Handler, error) { return newHandler(!b.DisableExporterMetrics, b.MaxRequests, b.Logger), nil }, needed?

Good question @roidelapluie and I realized this is not so clear on the PR description.

The reason behind this is because the final handler might depend on values that are only available after bootstrap has parsed the common flags and initialized logging.

In the node_exporter integration, the handler is built from `b.Logger`, `b.DisableExporterMetrics`, and `b.MaxRequests`, e.g.: https://github.com/prometheus/node_exporter/pull/3660

So the factory is the exporter-specific assembly point once the shared bootstrap state is resolved. Without it, callers would either have to duplicate bootstrap parsing/setup outside this package, or bootstrap itself would need to know how to construct exporter-specific handlers.

ArthurSens:
One thing that I just thought about!

If all CLI flags are registered through this bootstrapping package, we could create a [hidden](https://pkg.go.dev/github.com/alecthomas/kingpin/v2#ArgClause.Hidden) kingpin command to print them all.

I believe almost all exporters in prometheus and prometheus-community have their CLI flags in the README, but they are updated manually. We could easily build the automation needed to keep the README and the real world in sync

ArthurSens:
Taking the liberty to merge to unblock Nicolas!

PR #396, opened 2026-05-28
web: configurable IP TTL / hop-limit / DSCP on listening socket

randomizedcoder:
## Summary

Add three configurable IP-layer header fields on the exporter's listening socket — IPv4 TTL, IPv6 Hop Limit, and DSCP. Configuration via CLI flag, environment variable, or YAML `web-config.file`. Two motivations:

1. **Defense-in-depth via low TTL.** TTL=2 means scrape responses cannot escape past the second router hop, even if firewalls or ACLs ever fail open. The exporter becomes incapable of leaking metric data beyond the immediate L3 neighborhood.
2. **QoS via DSCP marking.** Operators on networks with traffic shaping / differentiated services can mark scrape traffic with a specific codepoint (CS2, AF11, EF, …) so intermediate equipment classifies it correctly.

All three knobs touch the IP header on the same socket via similar mechanisms, so they ship together. Every toolkit-using exporter (`node_exporter`, `blackbox_exporter`, `snmp_exporter`, …) gains the capability with a single change here.

## Configuration surface

| Knob | Flag | Env var | YAML field | Range |
|---|---|---|---|---|
| IPv4 TTL | `--web.ipv4-ttl` | `WEB_IPV4_TTL` | `ip_socket_config.ipv4_ttl` | 1–255 |
| IPv6 Hop Limit | `--web.ipv6-hop-limit` | `WEB_IPV6_HOP_LIMIT` | `ip_socket_config.ipv6_hop_limit` | 1–255 |
| DSCP | `--web.dscp` | `WEB_DSCP` | `ip_socket_config.dscp` | 0–63 |

Precedence: flag > env > YAML > kernel default. Omitting all sources leaves kernel defaults — full backwards compatibility.

## How it works

- **TTL / Hop Limit** are set on the *listening socket* via `net.ListenConfig.Control`. On Linux these options are inherited by accepted connections including the SYN-ACK packet (`accept(2)`, `ip(7)`, `ipv6(7)`).
- **DSCP** is NOT inherited (`IP_TOS` / `IPV6_TCLASS` reset on accept), so a thin `ipSocketListener` wraps the underlying `net.Listener` and applies the option to each accepted connection's FD via `(*net.TCPConn).SyscallConn()`. ECN bits in the ToS byte are deliberately left for the kernel to manage per-packet (RFC 3168).
- Three listener flavors handled:
  - Regular TCP: `net.ListenConfig.Control` + `ipSocketListener` wrap.
  - VSOCK: logs `"Ignoring IP socket options on VSOCK listener"` and proceeds unchanged (no IP layer).
  - Systemd socket activation: options applied post-bind via `(*net.TCPListener).SyscallConn()`; same wrapper added if DSCP is configured.

## Platform support

| Platform | Status |
|---|---|
| Linux | Fully supported; CI-tested. |
| FreeBSD / DragonFly / NetBSD / OpenBSD / Darwin | Compile-supported via `golang.org/x/sys/unix`; no CI run yet. |
| Windows / Plan 9 / JS+Wasm / others | No-op. First time any option is configured, a single warn-level log line is emitted and the values are ignored. |

## Commit organization

Each commit builds cleanly on its own (`go build`, `go vet`, `go test ./web/...`) and is independently revertable.

| # | Commit | Notes |
|---|---|---|
| 1 | `docs: add design brief` | Pre-impl design (`docs/ip-socket-config-design.md`). Drop if you don't want design docs in `docs/`. |
| 2 | `web: add IP socket-options config plumbing` | `FlagConfig` fields, `IPSocketConfig` YAML struct, `kingpinflag` registrations, validator. Pointer types so YAML-absent ≠ explicit-zero (load-bearing for DSCP=0, which is a valid configured value). |
| 3 | `web: apply IP socket options on the listening socket and accepted conns` | `applyListenerOptions` (TTL/Hop) + `applyConnOptions` (DSCP) — split because Linux inherits the former but not the latter — plus `ipSocketListener` wrapper and wiring in `ListenAndServe`. |
| 4 | `web: tests for IP socket options` | 5 new testdata YAMLs into existing `TestYAMLFiles`; Linux-gated `TestApplySocketOptions_Inheritance` with 9 subtests covering positive / boundary / corner cases; `TestEffective` for the precedence helper. |
| 5 | `web: docs and example for ip_socket_config` | `docs/web-configuration.md` section + commented example in `web/web-config.yml`. |
| 6 | **`[optional] web: refactor handler_test.go to table-driven`** | **Optional, labeled in commit subject.** Folds the 4 existing test functions in `handler_test.go` into one `TestHandler` table. No behavior change — same assertions, same outcomes. Justified on maintainability (adding a 5th test of the same shape becomes one row), not on enabling this feature. Drop via `git rebase -i` if you'd rather keep the existing structure; the feature commits before it are unaffected. |
| 7 | `web: validate DSCP flag value at resolve time` | Bug fix discovered during end-to-end testing: `kingpin.Int()` accepted `--web.dscp=999` and the kernel silently took the low byte. Adds a flag-level range check + test. Lands after commit 6 only because reordering would have needed an interactive rebase; **this commit should be kept regardless of whether commit 6 is dropped**. |

## Backwards compatibility

- New fields on `FlagConfig` are pointer-typed (nil by default). Existing struct literals that name only the old fields continue to compile.
- New YAML block `ip_socket_config:` is optional. Configs that don't include it behave identically to before.
- No behavior change for any existing flag/option.

## Test plan

- [x] `go test ./web/...` clean on Linux.
- [x] Cross-compile clean for `GOOS=darwin,freebsd,windows` (CGO disabled, as expected).
- [x] End-to-end smoke tests against `node_exporter` (built via `go mod replace` against this branch): flag-only, env-only, YAML-only, flag-overrides-YAML, validation rejects bad YAML (`ipv4_ttl: 0`, `dscp: 64`), validation rejects out-of-range flag (`--web.dscp=999`).

## Design doc

`docs/ip-socket-config-design.md` (commit 1) is the design brief — covers the rationale for the type asymmetry (`uint8` for TTL/Hop, `int` for DSCP because 0 is valid CS0), the inheritance asymmetry that drove the split between `applyListenerOptions` and `applyConnOptions`, the platform matrix, and the test coverage matrix. Includes an Appendix A specifically discussing the optional refactor commit and why it's separable.

Companion doc on the operator-facing side (in `node_exporter`): https://github.com/randomizedcoder/node_exporter/blob/ttl/docs/IP_SOCKET_CONFIG.md

### DIFF
--- a/docs/ip-socket-config-design.md
+++ b/docs/ip-socket-config-design.md
@@ -1,0 +1,402 @@
+# Design Brief: Configurable IP Socket Options (TTL, Hop Limit, DSCP) for `web.ListenAndServe`
+
+Status: Draft — pre-PR design brief
+Author: dave.seddon.ca@gmail.com
+Branch: `ttl` (fork: `randomizedcoder/exporter-toolkit`)
+Companion doc: [`node_exporter:docs/IP_SOCKET_CONFIG.md`](https://github.com/randomizedcoder/node_exporter/blob/ttl/docs/IP_SOCKET_CONFIG.md) (security/QoS rationale and exporter-side perspective)
+
+## Introduction
+
+This brief proposes adding configurable IP-layer header fields — IPv4 TTL, IPv6 Hop Limit, and DSCP (applied uniformly to IPv4 ToS and IPv6 Traffic Class) — to the listening sockets created by `github.com/prometheus/exporter-toolkit/web`. Two motivations:
+
+1. **Security via TTL clamping.** TTL=2 means packets die after two router hops; the exporter cannot leak metric data beyond the immediate L3 neighborhood even if firewalls or ACLs misbehave.
+2. **QoS via DSCP marking.** Operators on networks with traffic shaping or differentiated services need to classify scrape traffic with a specific codepoint (CS2, AF11, EF, etc.) so intermediate equipment queues it correctly.
+
+All three knobs touch the IP header on the same socket via the same `net.ListenConfig.Control` chokepoint with the same `setsockopt` pattern and the same Linux inheritance semantics on accepted connections. Bundling them in one feature is natural; landing it here propagates to every toolkit-using exporter.
+
+## Table of contents
+
+- [Introduction](#introduction)
+- [Background](#background)
+- [Configuration surface](#configuration-surface)
+- [Step 1 — Config plumbing](#step-1--config-plumbing)
+- [Step 2 — Socket-control function](#step-2--socket-control-function)
+- [Step 3 — Wire into `ListenAndServe`](#step-3--wire-into-listenandserve)
+- [Step 4 — Tests](#step-4--tests)
+- [Step 5 — Documentation](#step-5--documentation)
+- [Step 6 — Platform support and edge cases](#step-6--platform-support-and-edge-cases)
+- [Summary](#summary)
+
+## Background
+
+`ListenAndServe` in `web/tls_config.go` creates the TCP listener via a bare `net.Listen("tcp", address)` at line 323. Go's standard library exposes the listening socket's file descriptor before `bind(2)` via `net.ListenConfig.Control`; calling `unix.SetsockoptInt` at that point sets options on the listening socket. On Linux, the relevant options — `IP_TTL`, `IP_TOS`, `IPV6_UNICAST_HOPS`, `IPV6_TCLASS` — are all inherited by sockets returned from `accept(2)`, including the SYN-ACK packet (`accept(2)`, `ip(7)`, `ipv6(7)`). `tls.NewListener` is a transparent wrapper, so the settings flow through TLS unchanged.
+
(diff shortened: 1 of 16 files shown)


randomizedcoder:
Hi @SuperQ — thanks for maintaining exporter-toolkit!

I've rebased this onto the latest `master` and resolved the conflicts (they were confined to `go.mod` dep bumps and the `FlagConfig` / `checkFlags` refactor that landed with #394). While there I tidied the history: the DSCP-validation fix is now folded into the commit that introduces `resolveSocketOptions`, so all six commits build/vet/test independently. `go test ./web/...` is green locally on Linux.

A bit of real-world motivation: today we enforce the low-TTL behaviour out-of-band with `nftables` rules that rewrite the TTL on the exporter's scrape responses. It works, but it's fragile — the rules live outside the exporter, must be kept in sync with every deployment, and silently stop protecting us if they're ever flushed or the exporter moves host. An application-level knob set once on the listening socket is a much cleaner, more robust fit — and it benefits every toolkit-based exporter at once.

Would you have a chance to take a look, or point me at the right reviewer? Happy to adjust the approach or split anything out. The `[optional]` handler_test refactor commit is clearly labelled and can be dropped without touching the feature if you'd prefer. Thanks!

randomizedcoder:
@nicolastakashi Thanks for the review on https://github.com/prometheus/exporter-toolkit/pull/421 - I'd love your thoughts on the hop limits, which is an important security feature.

PR #417, opened 2026-07-16
web: add X25519MLKEM768 post-quantum curve to TLS config

Thealisyed:
## Summary

Add the X25519MLKEM768 hybrid post-quantum key exchange to the supported `curve_preferences` map in `web-config.yaml`.

## Motivation

X25519MLKEM768 (CurveID 4588) combines X25519 with ML-KEM-768 (FIPS 203). Go 1.24+ includes it in the default TLS 1.3 key exchange preferences when `CurvePreferences` is nil. However, users who explicitly set `curve_preferences` in their web config to control which key exchanges are offered cannot currently include `X25519MLKEM768`, which effectively disables post-quantum key exchange for those deployments.

This is particularly relevant for projects like Prometheus exporters running in environments with post-quantum cryptography mandates (e.g., OpenShift, which enforces centralized TLS security profiles).

## Changes

- `web/tls_config.go`: Add `X25519MLKEM768` to the `curves` map
- `web/testdata/web_config_noAuth_allCurves.good.yml`: Add to test fixture
- `docs/web-configuration.md`: Update description to cover non-ECDHE key exchanges

## Notes

- Go 1.26 will add `SecP256r1MLKEM768` and `SecP384r1MLKEM1024`. Those can be added in a follow-up when this module bumps its Go version requirement.
- `X25519MLKEM768` is TLS 1.3 only. If configured with `min_version: TLS12`, Go will silently skip it during TLS 1.2 handshakes — no error, it just won't be negotiated. This is the same behavior as the existing curves.

Signed-off-by: Ali Syed <alsyed@redhat.com>

### DIFF
--- a/docs/web-configuration.md
+++ b/docs/web-configuration.md
@@ -86,7 +86,7 @@ tls_server_config:
   # the order of elements in cipher_suites, is used.
   [ prefer_server_cipher_suites: <bool> | default = true ]
 
-  # Elliptic curves that will be used in an ECDHE handshake, in preference
+  # Elliptic curves and key exchange mechanisms that will be used in a TLS handshake, in preference
   # order. Available curves are documented in the go documentation:
   # https://golang.org/pkg/crypto/tls/#CurveID
   [ curve_preferences:
--- a/web/testdata/web_config_noAuth_allCurves.good.yml
+++ b/web/testdata/web_config_noAuth_allCurves.good.yml
@@ -7,3 +7,4 @@ tls_server_config:
     - CurveP384
     - CurveP521
     - X25519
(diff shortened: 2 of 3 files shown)


metalmatze (approved):
Makes sense to me.

metalmatze:
@Thealisyed, I can see you signed off on your commit, but something isn't quite right. Please have a look at the DCO CI run and check how to fix it. Thanks.

Thealisyed:
@metalmatze , thanks and is fixed now!

davidesalerno:
Is there something that it is blocking this PR to be merged?

PR #421, opened 2026-07-20
bootstrap: set ReadHeaderTimeout on the HTTP server

randomizedcoder:
Hi again 👋 — a small hardening follow-up, as mentioned in #420.

### Why

`Runner.newServer` in `bootstrap/` returns `&http.Server{Handler: mux}` with `ReadHeaderTimeout` unset, so request-header reads are unbounded — a Slowloris-style connection-exhaustion vector (gosec G112). Because the `web` package does not set server timeouts either, every bootstrap-based exporter inherits the gap.

This is the toolkit-side counterpart to prometheus/blackbox_exporter#1626, which set the same timeout on the caller side. Fixing it here addresses it once for all toolkit users.

### What

Add a `ReadHeaderTimeout time.Duration` field to `bootstrap.Config`, defaulting to one minute when unset. It bounds only header reading, not the metrics handler, so it never affects legitimate scrapes (which can run for the full scrape timeout), and — like every other `Config` field — downstream exporters can tune it to their own needs.

Table-driven tests assert the field maps onto the server (default when unset, configured value otherwise, non-positive treated as unset); a behavioral test confirms the server closes a connection whose headers never complete while a well-formed request still succeeds.

Thanks to @nicolastakashi for the excellent suggestion to make this a `Config` field rather than a package-level var — it's a much better fit for how the rest of `Config` is configured (plain struct fields, not functional options) and lets downstream exporters tune it too.

`go build`, `go vet`, `go test ./...` (incl. `-race`) pass; gosec G112 clears. Happy to adjust the value if a different bound is preferred.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -22,6 +22,7 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/prometheus/common/promslog"
@@ -42,6 +43,9 @@ var (
 	// errNegativeMaxRequests is returned when max requests is configured below zero.
 	errNegativeMaxRequests = errors.New("web max requests must be greater than or equal to zero")
 )
+
+// defaultReadHeaderTimeout applies when Config.ReadHeaderTimeout is left unset.
+const defaultReadHeaderTimeout = time.Minute
 
 // MetricsHandlerFactory builds an exporter-specific metrics handler after the
 // common toolkit flags have been parsed.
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.go
```diff
@@ -43,6 +44,10 @@ var (
 	errNegativeMaxRequests = errors.New("web max requests must be greater than or equal to zero")
 )
 
+// serverReadHeaderTimeout bounds request-header reads to mitigate Slowloris
+// (gosec G112). Package var so tests can shorten it.
+var serverReadHeaderTimeout = time.Minute
```
nicolastakashi: suggestion: Rather than a package-level var (only overridable by this package's own tests), I'd consider adding a ReadHeaderTimeout time.Duration field to Config, defaulting to time.Minute when unset, that matches how the rest of Config is configured (plain struct fields, not functional options) and lets downstream exporters tune it too.

randomizedcoder: @nicolastakashi Good idea!  Thanks

nicolastakashi (approved):

nicolastakashi:
@SuperQ LGTM, waiting for your final review.

ArthurSens (approved):
LGTM once the co-pilot nit is fixed :)

randomizedcoder:
@ArthurSens Thanks for the feedback.  Addressed

I'd also love a review on: https://github.com/prometheus/exporter-toolkit/pull/396

ArthurSens:
Thanks, I'll try to take a look at the linked issue once I find some time 😭

PR #430, opened 2026-08-18
bootstrap: allow exporters to add routes

nicolastakashi:
## Problem

`bootstrap` builds the exporter's HTTP mux itself: `Run()` registers the metrics handler at `--web.telemetry-path` and the landing page at `/`, and nothing else can be added to it.

That is enough for `node_exporter`, but it blocks exporters that serve more than those two endpoints. `postgres_exporter` is the immediate case:

- it serves the multi-target `/probe` endpoint, which is the documented way to scrape remote instances via `auth_modules`
- it exposes the `net/http/pprof` handlers

Adopting `bootstrap` as it stands would silently 404 both, so the exporter cannot move onto the shared startup path without a user-visible regression. Any exporter with a second endpoint hits the same wall, which pushes them back to hand-rolled `main()` blocks — the duplication `bootstrap` exists to remove.

A second constraint is *when* those handlers can be built. `postgres_exporter`'s probe handler needs the logger and the loaded config file, neither of which exists until flags are parsed, so a static list of routes on `Config` is not sufficient.

## Follow-up

Consumed by prometheus-community/postgres_exporter#1368, which moves `postgres_exporter` onto `bootstrap` while keeping `/probe` and pprof working. That PR needs a release of this one before it builds.

Note: an earlier revision of this PR also let exporters back `--web.telemetry-path` with a custom env var, to preserve `postgres_exporter`'s `PG_EXPORTER_WEB_TELEMETRY_PATH`. Per @ArthurSens's review, that's dropped — new env-var-backed flags shouldn't be added, existing ones should be deprecated/removed instead. The follow-up postgres_exporter PR will need to handle that separately.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -41,9 +41,10 @@ var (
 	errEmptyMetricsPath = errors.New("metrics path must not be empty")
 	// errNegativeMaxRequests is returned when max requests is configured below zero.
 	errNegativeMaxRequests = errors.New("web max requests must be greater than or equal to zero")
-	// errProtectedRoutePattern is returned when a caller-registered route
-	// conflicts with a path bootstrap manages itself.
-	errProtectedRoutePattern = errors.New("route pattern is reserved by bootstrap")
+	// errReservedMetricsPath is returned when a caller-registered route
+	// conflicts with --web.telemetry-path. The root path may be overridden
+	// by a caller-registered route, but the metrics path may not.
+	errReservedMetricsPath = errors.New("route pattern is reserved for the metrics handler")
 )
 
 // MetricsHandlerFactory builds an exporter-specific metrics handler after the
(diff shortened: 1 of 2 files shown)


ArthurSens (changes requested):
I don't think we want to continue using env vars in exporters. It would be better to deprecate and remove this functionality from existing exporters rather than continue building on top of it

nicolastakashi:
> I don't think we want to continue using env vars in exporters. It would be better to deprecate and remove this functionality from existing exporters rather than continue building on top of it

Remove the env var support in here.

ArthurSens (commented):
What happens if someone adds a new route that conflicts with `/` or `/metrics`? Should we provide guardrails against that? 

What ever we decide, it would also be nice to have a test that confirms the behavior we want

ArthurSens:
Could update your node/postgres PRs with a replace in go mod, just so we can see how these changes look like downstream before we merge?

nicolastakashi:
> What happens if someone adds a new route that conflicts with `/` or `/metrics`? Should we provide guardrails against that?
> 
> What ever we decide, it would also be nice to have a test that confirms the behavior we want

Good suggestion, I'll take it.

ArthurSens:
Sorry for the back-and-forth here, but after thinking a bit more, it feels like we shouldn't reserve the bootstrap routes. I can imagine that tons of exporters already have something running at the `/` route, for example. Prometheus is one example, where the landing page redirects to `/query`.

Maybe we should allow overrides, and our test would focus on making sure the override takes precedence over the routes that bootstrap creates?

ArthurSens:
Hey @nicolastakashi, not sure if you requested my review by mistake, but my last comment is still unaddressed 😅

nicolastakashi:
> Sorry for the back-and-forth here, but after thinking a bit more, it feels like we shouldn't reserve the bootstrap routes. I can imagine that tons of exporters already have something running at the `/` route, for example. Prometheus is one example, where the landing page redirects to `/query`.
> 
> Maybe we should allow overrides, and our test would focus on making sure the override takes precedence over the routes that bootstrap creates?

`/metrics` stays reserved (errors on conflict), but a caller route at / now overrides the landing page.

@ArthurSens PTAL 🙏🏽

ArthurSens (approved):
LGTM

PR #434, opened 2026-09-01
web: emit systemd sd_notify readiness notification on startup (#389)

### LINKED ISSUE #389: feature: emit systemd sd_notify readiness notification on startup
### Summary

`exporter-toolkit` already supports systemd socket activation (added in #95), using `go-systemd/v22/activation`. It would be a natural
complement to also send a `READY=1` readiness notification to systemd
once the exporter is actually listening and ready to serve requests.

### Motivation

Without `sd_notify` support, exporters built on this toolkit cannot use
`Type=notify` in their systemd service units. They are limited to
`Type=simple` or `Type=exec`, where systemd considers the service
"started" as soon as the process is exec'd — before it is actually
listening on its port.

This matters in practice when:

- Other units declare `After=<exporter>.service` and need the exporter
  to be truly ready before they proceed.
- Operators want accurate startup ordering and timing in `systemctl
  status` / `systemd-analyze`.
- The exporter starts slowly (e.g. expensive collector initialisation)
  and a dependent unit races against it.

Since all exporters using this toolkit share the same `ListenAndServe`
path, implementing this once here would benefit the entire ecosystem
(node_exporter, blackbox_exporter, etc.) for free.

### Prior art in this repo

- #95 / #225: socket activation support and the open request to
  auto-detect `$LISTEN_FDS` — this feature is in the same spirit of
  "be a well-behaved systemd service automatically".
- #152: open issue on signal handling for proper systemd unit operation
  — further evidence that systemd integration is a recurring concern.

### Implementation notes

The relevant API is `github.com/coreos/go-systemd/v22/daemon`, specifically:

```go
daemon.SdNotify(false, daemon.SdNotifyReady)
```

This module is **already a direct dependency** of `exporter-toolkit`
(used by the `activation` sub-package for socket activation), so no new
dependency would be introduced.

### Expected outcome

Exporters using this toolkit would work correctly with
`Type=notify` in their systemd service unit:

```ini
[Service]
Type=notify
ExecStart=/usr/bin/node_exporter
```

Systemd would then only consider the service "active (running)" once the
exporter signals readiness, enabling reliable `After=` ordering for
dependent units.

SHIVANSH-ux-ys:
Fixes #389

Emit systemd readiness notification (`daemon.SdNotify(false, daemon.SdNotifyReady)`) in `ServeMultiple` when server listeners start serving requests. This enables exporters built with `exporter-toolkit` to work correctly with `Type=notify` in systemd units.

### DIFF
--- a/web/tls_config.go
+++ b/web/tls_config.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/coreos/go-systemd/v22/activation"
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/mdlayher/vsock"
 	config_util "github.com/prometheus/common/config"
 	"go.yaml.in/yaml/v2"
@@ -302,6 +303,7 @@ func ConfigToTLSConfig(c *TLSConfig) (*tls.Config, error) {
 // ServeMultiple starts the server on the given listeners. The FlagConfig is
 // also passed on to Serve.
 func ServeMultiple(listeners []net.Listener, server *http.Server, flags *FlagConfig, logger *slog.Logger) error {
+	_, _ = daemon.SdNotify(false, daemon.SdNotifyReady)
 	errs := new(errgroup.Group)
 	for _, l := range listeners {
 		errs.Go(func() error {


web/tls_config.go
```diff
@@ -309,6 +303,7 @@ func ConfigToTLSConfig(c *TLSConfig) (*tls.Config, error) {
 // ServeMultiple starts the server on the given listeners. The FlagConfig is
 // also passed on to Serve.
 func ServeMultiple(listeners []net.Listener, server *http.Server, flags *FlagConfig, logger *slog.Logger) error {
+	_, _ = daemon.SdNotify(false, daemon.SdNotifyReady)
```
ArthurSens, line 306: We should notify ready at the end, once everything (config, auth, certificates, TLS) was verified to run correctly

ArthurSens: A test verifying these things would also be nice:

* READY!=1 when initialization fails
* READY=1 when initialization succeeds

ArthurSens (changes requested):

PR #435, opened 2026-09-02
web: fix nil pointer dereference when systemd socket activation is disabled

mrueg:
`checkFlags` only required listen addresses to be set when `WebSystemdSocket` was `nil`:

```go
if c.WebSystemdSocket == nil && (c.WebListenAddresses == nil || len(*c.WebListenAddresses) == 0) {
	return ErrNoListeners
}
```

A non-nil pointer to `false` satisfies that condition, so a `FlagConfig` with systemd socket activation explicitly *disabled* and no listen addresses passes validation. `ListenAndServe` then falls through to the port-listener path and panics dereferencing the nil pointer:

```go
listeners := make([]net.Listener, 0, len(*flags.WebListenAddresses))
```

`kingpinflag.AddFlags` always hands out a non-nil `WebSystemdSocket` pointer (it points at `false` on non-Linux), so this is reachable from any caller that builds a `FlagConfig` directly instead of going through the kingpin flags — `FlagConfig` is exported with documented fields, so that is a supported use.

Fix: require listen addresses whenever systemd socket activation is not actually enabled.

`checkFlags`, `ErrNoListeners` and `ErrMissingFlag` had no test coverage at all, so this adds a table test for `checkFlags` plus an end-to-end check that `ListenAndServe` returns `ErrNoListeners` rather than panicking. Reverting the one-line fix makes the new tests fail with the nil dereference panic.

### DIFF
--- a/web/flags_test.go
+++ b/web/flags_test.go
@@ -1,0 +1,102 @@
+// Copyright 2025 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


ArthurSens (approved):
LGTM, thanks!

PR #444, opened 2026-09-02
web: mark prefer_server_cipher_suites as deprecated and ignored

mrueg:
The documentation gives `prefer_server_cipher_suites` a four-line explanation of how it changes cipher selection, and it has not done anything for years.

It is passed to `tls.Config.PreferServerCipherSuites`, whose own documentation reads:

> PreferServerCipherSuites is a legacy field and has no effect.

Go has selected the cipher suite itself since Go 1.17. Operators tuning this are tuning a knob that is not connected to anything.

### Change

Say so in the documentation and on the struct field, and stop passing the value to `tls.Config` — which also removes a use of a deprecated standard library field. The YAML key is still parsed and still defaults to `true`, so existing configuration files keep loading unchanged.

### DIFF
--- a/docs/web-configuration.md
+++ b/docs/web-configuration.md
@@ -80,10 +80,13 @@ tls_server_config:
   [ cipher_suites:
     [ - <string> ] ]
 
-  # prefer_server_cipher_suites controls whether the server selects the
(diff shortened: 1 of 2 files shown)


ArthurSens (approved):
Thanks!

PR #447, opened 2026-09-18
bootstrap: apply --web.max-requests, allow opting routes in

nicolastakashi:
--web.max-requests is parsed but never enforced, and routes like postgres_exporter's `/probe` get no concurrency protection at all (see prometheus-community/postgres_exporter#1368).

This binds the metrics endpoint to the flag, and adds `Bootstrap.MaxRequestsHandler` so other scrape-shaped routes can opt in too. Health checks stay unbounded by default.


bootstrap/bootstrap.go
```diff
@@ -64,14 +66,53 @@ type Bootstrap struct {
 	MetricsPath string
 	// FlagConfig contains the parsed exporter-toolkit web flags.
 	FlagConfig *web.FlagConfig
-	// DisableExporterMetrics reports whether exporter self-metrics should be disabled.
+	// DisableExporterMetrics is the parsed value of --web.disable-exporter-metrics.
```
ArthurSens, line 69: I know this is out of scope for this PR, but I'd love to see exporter-toolkit owning a Registry and this boolean would handle registering VersionCollector, GoCollector and HTTP metrics in the future :)

It sounds a bit weird to have this boolean here and then have an exporter miss it by accident.

ArthurSens (approved):
LGTM!

